PR #49, opened 2013-02-06
MONDRIAN-1272: Support for empty cells in native top count

pamval:
Added capability to fill in the result set with additional members if top count returns less than the limit provided (only if non empty is false).

I'm using a new constraint (MemberExcludeConstraint) to filter out the members returned by the top count. Also had to do some role enforcing there - not sure that's the right place to do it, so suggestions welcome.


lucboudreau:
Replaced with PR #566

PR #61, opened 2013-03-27
Convert Mondrian to use SLF4J.

magnayn:
I hit MONDRIAN-1171, and I wasn't able to patch the relevant 3rd party JARs, so I thought I'd fix Mondrian to use slf4j natively.

This is a work in progress commit for comment; - I suspect the tests don't work as I can't run them. It shouldn't be much work to make them work. If you can give me some pointers to how to run them, I can adjust the patch.


magnayn:
Is this project dead?

It appears you want to use SLF4J (which makes sense) and this patch basically does (did), yet it sits here for a couple of years??

lucboudreau:
https://jira.pentaho.com/browse/MONDRIAN-1508

magnayn:
Still proudly bound to a private log4j API I see, 6 years later.
😂 😂 😂

PR #62, opened 2013-03-28
Mavenize

magnayn:
The POM file that gets generated for Mondrian is very junky. e.g: contains very many dependency declarations that are not necessary. 

The easiest way to fix all of that is to make it actually build with maven.

This patch series is a 'starter for 10'. It creates a plugin to do the properties generation stuff, and separates out the required exception classes to compile in a different project.

It generates a pom.xml that uses this - deferring to ANT scripts from the original build file where there isn't a plugin (e.g: resgen, xomgen). dependency:analyze was used to remove un-needed deps (though it may have this wrong for tests).

It does not touch the original files, so as-is, the ant build still works.

Many tests seem to be erroring. This could be OK (version mismatches), but there's some wierd GC overhead stuff elsewhere.

Further things to do :
- generate plugins for res / xom, possibly within their respective projects.
- further split apart the project. It looks to me like (e.g) CmdRunner ought to be a part of a different project (that perhaps can then use Logback directly, solving the slf4j patch issue). Remove classes only used in the build process from the jar.
  - reorganise (for the above) into more maven-style directory structures. Though this would break the ant build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,7 @@ src/main/mondrian/rolap/aggmatcher/aggregates.dtd
 src/main/mondrian/xmla/DataSourcesConfig.java
 src/main/mondrian/xmla/datasourcesconfig.dtd
 
+.idea
+target
+*.iml
+bin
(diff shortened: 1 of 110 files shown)


lucboudreau:
Maven conversion was done as of 8.0.

PR #227, opened 2014-02-13
Fix for log4j.jar not ending up in ${lib.dir}

frankgrimes97:
http://stackoverflow.com/questions/6218454/ivy-not-installing-some-jars


lucboudreau:
We moved to Maven.

PR #256, opened 2014-04-08
[MONDRIAN-1910] NPE in XmlaHandler with empty axis

tgf:
http://jira.pentaho.com/browse/MONDRIAN-1910
Axes with empty sets cause NPE in XmlaHandler.


lucboudreau:
Fixed in #922.

PR #310, opened 2014-05-14
Drain the eventQueue on server shutdown

sweeze:
to prevent executor threads from waiting forever for a response that will never arrive.


lucboudreau:
https://jira.pentaho.com/browse/MONDRIAN-2625

PR #371, opened 2014-08-06
Fix log4j method not found and commons-io not in WEB-INF/lib

goncha:
Fix log4j method not found and commons-io not in WEB-INF/lib
- [derby-]war tasks use log4j-1.2.8.jar from jpivot.war to build war file
- [derby-]war tasks don't copy commons-io.jar into dest war file
- This fix uses all dependencies in mondrian to overwrite jars from jpivot.war


lucboudreau:
We moved to Maven.

PR #411, opened 2014-12-19
[MONDRIAN-2202] Addresses issues with native evaluation and with

mkambol:
CJFD.nonEmptyList that arise when one or more measures change the
conditions under which a context is non empty.  For example, given a
measure which sums YTD values up to Time.CurrentMember, it's possible
that there will be non empty data for tuples in a crossjoin even if
there is no base measure value at Time.CurrentMember (since the
measures sums *up to* CurrentMember).

This fix extracts all referenced members from the set of measures in
the query, and determines whether there's a potential for conflict.
If there is, native eval is turned off.  Likewise a conflict in
CJFD.nonEmptyList is handled by removing the conflicting member from
the evaluation context to avoid suppressing tuples which may
ultimately have data.

This change does not address calculated members not defined on
measures that result in similar changing of context.  It's possible
that the approach can be extended to handle calc members in general,
but the challenge is to avoid disabling native evaluation too broadly.


src/main/mondrian/olap/fun/MemberExtractingVisitor.java
```diff
@@ -0,0 +1,144 @@
...
+    private final ResolvedFunCall call;
+    private final boolean mapToAllMember;
+
+    /**
+     * This list of functions are "blacklisted" because
+     * occurrence of a member/dim/level/hierarchy) within
+     * one of these expressions would result in a set of members
+     * that cannot be determined from the expression itself.
+     */
+    private static final String[] unsafeFuncNames = new String[] {
```
lucboudreau: Should 'Ancestor(s)' be in that list as well?

mkambol: I think you're right- it should be.  It would be worth a full review of the function list before merging.  There may be others omitted.

julianhyde: Forgive me if I've said this already: There's a reasonable chance that the "unsafe" functions can be detected by their "depends on" metadata. If so this list would be wholly or partly redundant.

src/main/mondrian/olap/fun/MemberExtractingVisitor.java
```diff
@@ -0,0 +1,144 @@
...
+import mondrian.olap.*;
+import mondrian.olap.type.Type;
+
+import java.util.*;
+
+/**
+ * Visitor which collects any non-measure base members encountered while
+ * traversing an expression.
+ *
+ * This Visitor is used by the native set classes as well as the crossjoin
```
julianhyde: Need <p> paragraph markers

src/main/mondrian/olap/fun/MemberExtractingVisitor.java
mkambol: I explored use of dependsOn(), but found many cases where it wasn't helpful to identify possible conflict.  For example, with a crossjoin of customers and products, if a measure on an opposing axis computes YTD sales, then [Time] is potentially conflicting, but will not be dependent.  Cases where the potentially conflicting member is directly in the crossjoin also may show no dependence.  For example if Gender.[F] is in the CJ it will be treated as a constant calc with dependence=false, even though it may be overridden elsewhere to be Gender.[M].

If you think there are other things worth trying please let me know.  Use of the unsafe list is pretty ugly but I couldn't come up with a better alternative.

PR #412, opened 2014-12-19
MONDRIAN-2161 check cancel during sql fetch

tgf:
@lucboudreau please review


tgf:
squashed in
https://github.com/pentaho/mondrian/pull/414

PR #425, opened 2015-02-02
[MONDRIAN-2260] Fixes an issue where a compound slicer used with a share...

lucboudreau:
...d dimension within a virtual cube with a context of joining and non-joining dimensions would throw a null pointer exception. Per the API, if the request cannot be satisfied as-is, we must return null directly.


lucboudreau:
Waiting on ad-hoc triage before merging.

PR #426, opened 2015-02-10
Fix MONDRIAN-1371.

avillar:
This patch fixes MONDRIAN-1371, by setting OlapConnection's locale if the client sent any.


lucboudreau:
Fixed by #945

PR #429, opened 2015-03-02
SP-1720: Backport of MONDRIAN-2220 - Connecting to Mondrian via XMLA is creating invalid SQL against Sybase ASE via jConnect JDBC driver (5.3 Suite)

YuryBY:
@lucboudreau, would you mind reviewing/merging

Sybase dialect processing is not removed but improved according to syntax of convert() function.
Junit test is also provided.

### DIFF
--- a/src/main/mondrian/rolap/RolapMemberBase.java
+++ b/src/main/mondrian/rolap/RolapMemberBase.java
@@ -8,7 +8,6 @@
 // Copyright (C) 2005-2013 Pentaho and others
 // All Rights Reserved.
 */
-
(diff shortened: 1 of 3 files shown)


lucboudreau:
This is not an appropriate solution. All of that logic must be refactored into the dialect system. You might have to create a new method in Dialect, JdbcDialectImpl and SybaseDialect. I suspect that using 'convert' is not required anymore and you'll need to test on Sybase to figure this out.

Additionally, generating SQL by invoking toString() on a List is a no-no.

PR #435, opened 2015-03-11
[ANALYZER-2655] Invalid Agg table SQL generated by native filter when

mkambol:
the filter condition excludes measures and the level has extra columns
(properties, ordinalColumn, etc.)

Addresses a couple of cases mishandled by SqlTupleReader when joining
collapsed attributes which have additional columns that need to be
read from the dimension table(s).  For example, there may be need to
read ordinal column, or property values.  The existing logic worked
well enough in most cases, but made two faulty assumptions:

1) It assumed the aggstar will be joined in via the constraint, which
may not always be the case.
2) It assumed using the dimensional attribute in the select list
instead of the agg star attribute wouldn't have an impact.  This
normally shouldn't matter--the collapsed attribute and the
corresponding attribute from the dim table are identical-- but native
filter expects the filter constraint column from the aggstar to be
present.

### DIFF
--- a/bin/checkFile.awk
+++ b/bin/checkFile.awk
@@ -330,7 +330,7 @@ FNR < headerCount {
 /^\/\/ Copyright .* Pentaho/ && strict > 1 {
     # We assume that '--strict' is only invoked on files currently being
     # edited. Therefore we would expect the copyright to be current.
-    if ($0 !~ /-2014/) {
+    if ($0 !~ /-2015/) {
         error(fname, FNR, "copyright is not current");
     }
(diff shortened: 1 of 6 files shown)


mkambol:
@lucboudreau

lucboudreau:
Can you run this on Oracle & Postgres as well just out of curiosity? MySQL tends to be more permissive with columns being grouped and filtered even though they are not part of the select clause, versus Oracle which enforces a more strict standard. Just want to make sure that the new SQL works well in other DBs.

PR #441, opened 2015-03-20
compute column cardinality faster on PostgreSQL

rmwpl:
in PgSQL `count(*) from (select distinct)` is much faster, than `count(distinct)`


deinspanjer:
Is there ever a case where you need to get cardinality counts for multiple columns in the same table?
There is a significant advantage for the count(distinct) function in that case:
select count(distinct country), count(distinct region), count(distinct city) from geography;
versus
select count(*) from (select distinct country from geography);
select count(*) from (select distinct region from geography);
select count(*) from (select distinct city from geography);

rmwpl:
Damn, you're right, didn't thought about that, but it seems to me that Mondrian is actually issuing one query for each column using count(distinct)

PR #447, opened 2015-04-02
MONDRIAN-2259 - Sybase SQL Anywhere 16 - Incorrect syntax near 'DATE' shown in Analyzer

YuryBY:
@lucboudreau, would you mind reviewing/merging

### DIFF
--- a/src/main/mondrian/spi/impl/SybaseDialect.java
+++ b/src/main/mondrian/spi/impl/SybaseDialect.java
@@ -1,15 +1,18 @@
 /*
-* This software is subject to the terms of the Eclipse Public License v1.0
-* Agreement, available at the following URL:
-* http://www.eclipse.org/legal/epl-v10.html.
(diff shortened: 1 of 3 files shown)


lucboudreau:
@mkambol Please review & merge. Thanks!

mtseu:
@mkambol as far as I can see the validation is not bypassed by the fix.
date format check is performed in https://github.com/pentaho/mondrian/blob/master/src/main/mondrian/spi/impl/JdbcDialectImpl.java#L421
and then overridden method is called for a particular dialect:
https://github.com/YuryBY/mondrian/blob/master/src/main/mondrian/spi/impl/SybaseDialect.java#L53

So the change looks quite right.

mkambol:
Ah, you're right @mtseu, my mistake.  When I looked at this earlier I misread and thought it was an override of the public method.

PR #448, opened 2015-04-03
add a dialect for sparksql

Jeffrio:
recently,we use spark sql
there is no SparkSqlDialect supplied by mondiran
I bulid the spark sql accroding to the hiveDialect, and this work for me , now I get the the spark sql running on saiku with mondrian3.9

### DIFF
--- a/src/main/mondrian/olap/Mondrian_SW.xml
+++ b/src/main/mondrian/olap/Mondrian_SW.xml
@@ -1418,6 +1418,7 @@
             <Value>greenplum</Value>
             <Value>vectorwise</Value>
             <Value>hive</Value>
+            <Value>sparksql</Value>
         </Attribute>
         <CData/>
         <Code><![CDATA[
--- a/src/main/mondrian/spi/impl/SparkSqlDialect.java
+++ b/src/main/mondrian/spi/impl/SparkSqlDialect.java
@@ -1,0 +1,176 @@
+/*
+// This software is subject to the terms of the Eclipse Public License v1.0
+// Agreement, available at the following URL:
+// http://www.eclipse.org/legal/epl-v10.html.
+// You must accept the terms of that agreement to use this software.
+//
+// Copyright (C) 2011-2014 Pentaho and others
+// All Rights Reserved.
+*/
+package mondrian.spi.impl;
+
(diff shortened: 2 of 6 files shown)


src/main/mondrian/olap/Mondrian_SW.xml
```diff
@@ -1418,6 +1418,7 @@
             <Value>greenplum</Value>
             <Value>vectorwise</Value>
             <Value>hive</Value>
+            <Value>sparksql</Value>
```
lucboudreau: You should also update Mondrian.xml. This one here is for the schema workbench.

src/main/mondrian/spi/impl/SparkSqlDialect.java
```diff
@@ -0,0 +1,176 @@
...
+     *
+     * @param connection Connection
+     *
+     * @throws java.sql.SQLException on error
+     */
+    public SparkSqlDialect(Connection connection) throws SQLException {
+        super(connection);
+    }
+
+    protected String deduceIdentifierQuoteString(
```
lucboudreau: No quotes at all? This won't work with drillthrough operations, where the column alias might be two words. Not quoting them will create invalid SQL.

Jeffrio: spark sql use back-tick as quote

src/main/mondrian/olap/Mondrian_SW.xml
Jeffrio: I have updated  Mondrian.xml add sparksql

lucboudreau:
Thank you for this contribution. I've created this ticket to track it and our PMs will prioritize this feature.

http://jira.pentaho.com/browse/MONDRIAN-2328

SPECUSA:
Hello @Jeffrio 

I'm trying to add dialect for sparksql in mondrian. I performed the steps as you have shown to add dialect. But it work well in Pentaho CE though it doesn't work in Schema Workbench

I'm using Pentaho CE -5.0.1, Schema Workbench  3.9.0.0-213 and made changes in Mondrian 3.9

Please let me know why I can't get Sparksql dialect on schema workbench

Thanks in advance

PR #462, opened 2015-04-23
[BACKLOG-2384] Changes to better support loading and caching of

mkambol:
dimension members.

1) Introduces a "batch resolution" phase prior to final expression
resolution which collects child name references and loads them at the
same time.  This avoids the one-at-a-time SQL queries for each child
member reference within an expression.  IdBatchResolver is the
relevant class.

2) To better suppport (1), this change also introduces a new caching
mechanism for ChildByNameConstraints.  If the child members { [Bob],
[Joe], [Bill] } have already been cached, MemberCacheHelper is now
smart enough to retrieve any subset of what has formerly been loaded.
This new functionanlity only works with ChildByNameConstraint checks,
since it depends on lookups by name.

3) To better handle small levels (under 300 members), this change also
adds a check for level cardinality when constructing a constraint in
SqlConstraintFactory.  In general, any level with fewer than 300
members will be loaded with a default constraint to avoid repeat
queries to the dimension table.  There are some exceptions to this,
for example degenerate dimensions are assumed to be too expensive for
full table scans, so a Sql constraint is preferred.  This is governed by
a new property called "LevelPreCacheThreshold", which can be set to
0 to disable the new behavior.

### DIFF
--- a/src/main/mondrian/olap/DelegatingSchemaReader.java
+++ b/src/main/mondrian/olap/DelegatingSchemaReader.java
@@ -6,10 +6,9 @@
 //
 // Copyright (C) 2003-2005 Julian Hyde
 // Copyright (C) 2004-2005 TONBELLER AG
-// Copyright (C) 2005-2012 Pentaho
+// Copyright (C) 2005-2015 Pentaho
 // All Rights Reserved.
 */
-
 package mondrian.olap;
 
 import mondrian.calc.Calc;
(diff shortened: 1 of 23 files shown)


mkambol:
@lucboudreau

lucboudreau:
Would you mind also announcing this change to the public mailing list please? I anticipate a lot of questions and would prefer we address this up front. Thanks!

PR #466, opened 2015-04-28
[BACKLOG-2308] - introducing method isParentChildPhysicalMember to speci...

kurtwalker:
...fically identify them for reading level members.

### DIFF
--- a/src/main/mondrian/olap/Member.java
+++ b/src/main/mondrian/olap/Member.java
@@ -5,10 +5,9 @@
 // You must accept the terms of that agreement to use this software.
 //
 // Copyright (C) 1999-2005 Julian Hyde
-// Copyright (C) 2005-2010 Pentaho and others
+// Copyright (C) 2005-2015 Pentaho and others
 // All Rights Reserved.
 */
-
 package mondrian.olap;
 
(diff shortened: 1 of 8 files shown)


kurtwalker:
@lucboudreau or @mkambol you'll want to review this one

src/main/mondrian/olap/MemberBase.java
```diff
@@ -280,6 +279,10 @@ public String getPropertyFormattedValue(String propertyName) {
         return getPropertyValue(propertyName).toString();
     }
 
+    public boolean isParentChildPhysicalMember() {
```
lucboudreau: Probably worth adding javadocs here since I'm myself not sure what it means.

src/main/mondrian/olap/Member.java
```diff
@@ -63,6 +62,8 @@
 
     boolean isParentChildLeaf();
 
+    boolean isParentChildPhysicalMember();
```
lucboudreau: Actually, please add javadocs here instead. :)

kurtwalker:
@lucboudreau java doc added and commits squashed.

PR #467, opened 2015-04-29
Backport of MONDRIAN-2338 - REGRESSION: XMLA servlet does not recognize the Datasource Name in its legacy form (5.3 Suite)

YuryBY:
@lucboudreau, @kurtwalker here are changes according to Luck's patch


kurtwalker:
@lucboudreau I'm not familiar with this one, so I'll leave it for you.

PR #477, opened 2015-05-16
Update documentation re: 'all' member default name

karlwbrown:
Updating documentation to describe the convention of adding "s" to the end of default allMember names. (source code found in mondrian/src/main/mondrian/rolap/RolapHierarchy.java)

This patch came about as a result of some confusion I had with this convention, and I realized it wasn't documented.

http://forums.pentaho.com/showthread.php?186665-Getting-dimension-values-including-%28All%29-member

Please let me know if this patch is ok, and if you can get it up on the pentaho website so others can see it :)


karlwbrown:
This is a minor documentation fix outstanding for over 2 years - any word from the maintainers on this?

PR #483, opened 2015-06-10

akhayrutdinov:
- change Interval.diff() for D and Y members to mimic DateDiff function from VBA
- add test cases for DateDiff('y'/'d',...)
- do some cosmetic changes to make affected class follow the checkstyle

### DIFF
--- a/doc/vba_functions.html
+++ b/doc/vba_functions.html
@@ -500,7 +500,7 @@ functions implemented, is described in the <a href="mdx.html">MDX specification<
 	q Quarter <br>
 	m Month <br>
 	y Day of year <br>
-	d Day <br>
+	d Day of year<br>
 	w Weekday <br>
 	ww Week <br>
 	h Hour <br>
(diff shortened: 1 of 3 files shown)


doc/vba_functions.html
```diff
@@ -500,7 +500,7 @@
 	q Quarter <br>
 	m Month <br>
 	y Day of year <br>
```
lucboudreau: y = Day of year?

akhayrutdinov: @lucboudreau, yes it is. See it here: https://msdn.microsoft.com/en-us/library/b5xbyt6f%28v=vs.90%29.aspx

lucboudreau:
Just one more question to be sure. I'm asking because a similar situation happened in the past. Was this code taken as a copy from anywhere, or is this original work? I'm merely asking because date-related code tends to be copied around a lot and we must make sure that nothing gets copy-pasted in ours.

akhayrutdinov:
@lucboudreau, I adapted ```computeJdn()``` from here: https://github.com/pentaho/pentaho-vertica-bulkloader/blob/master/src/org/pentaho/di/verticabulkload/nativebinary/ColumnSpec.java#L328
I added there previously while solving another date-related problem. The formula is taken from Wikipedia, it is public domain :)

lucboudreau:
All content on Wikipedia, albeit 'public domain', is released according to the terms of a specific license. 

Let's merge it for now, but I'll check with our IP team. Great job btw.

julianhyde:
I can't believe that I did this wrong (and it's been wrong for all this time!). Thanks for fixing.

If you need some clean code you can copy from Calcite. Apache license, wrote it myself. DateTimeUtils.ymdToJulian.

https://github.com/apache/incubator-calcite/blob/94752a6066cd8a379782510a78a885cffb5aadb9/avatica/src/main/java/org/apache/calcite/avatica/util/DateTimeUtils.java#L708

akhayrutdinov:
@lucboudreau, @mkambol, do I need to add something more? The PR is neither merged, nor declined :)

lucboudreau:
This change broke the build. Please fix ASAP.

PR #484, opened 2015-06-16

akhayrutdinov:
- remove TimeUnit to make the class be compilable with JDK 1.5

@lucboudreau, @mkambol, merge it please

### DIFF
--- a/src/main/mondrian/olap/fun/vba/Vba.java
+++ b/src/main/mondrian/olap/fun/vba/Vba.java
@@ -16,7 +16,6 @@ import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import static java.util.concurrent.TimeUnit.*;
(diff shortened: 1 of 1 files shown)


julianhyde:
You are using int constants (e.g. 1000 * 60). To prevent overflow, use long constants (e.g. 1000L * 60).

akhayrutdinov:
@julianhyde, thanks for your comment. I've updated the PR.
However, overflow was imposible there (I had taken that into account previously):
```
max({(1000 * 60 * 60) * (hAfter - hBefore)}) = 1000 * 60 * 60 * 47 = 169200000 < Integer.MAX_VALUE = 2147483647
max({(1000 * 60) * (after.get(Calendar.MINUTE) - before.get(Calendar.MINUTE))}) = 1000 * 60 * 60 << Integer.MAX_VALUE 
```

PR #487, opened 2015-06-18
Mondrian-2251 - Exceeding CellBatchSize when evaluating a UDF causes an error

YuryBY:
@lucboudreau, @mkambol, would you mind of reviewing/merging


mkambol:
@YuryBY - did you run 
   bin/checkFile.sh --opened 
before committing?  Make sure to do that, or run checkFile.sh --strict on each file you've modified.

YuryBY:
@lucboudreau, @mkambol, I had run checkFile.sh without "--strict" parameter, already updated

PR #488, opened 2015-06-19
[MONDRIAN-2319] - Calculating Date difference produce wrong result

akhayrutdinov:
- fix a mistake with parameters' order

@lucboudreau, @mkambol, review it please.
I made a ridiculous mistake by confusing first and second parameters. Because of this, results are opposite to expected

PR #491, opened 2015-06-23
[SP-1949] - Backport of MONDRIAN-2319 - Calculating Date difference p…

akhayrutdinov:
…roduce wrong result (5.4 Suite)

* cherry picked from commit 1ee1157
 * change Interval.diff() for D and Y members to mimic DateDiff function from VBA
 * add test cases for DateDiff('y'/'d',...)
 * do some cosmetic changes to make affected class follow the checkstyle
* cherry picked from commit fe844dd
 * remove TimeUnit to make the class be compilable with JDK 1.5
* cherry picked from commit 95e7929
 * fix a mistake with parameters' order

@lucboudreau, @mkambol, review it please. This is a backport of these PRs: https://github.com/pentaho/mondrian/pull/483, https://github.com/pentaho/mondrian/pull/484, https://github.com/pentaho/mondrian/pull/488

### DIFF
--- a/src/main/mondrian/olap/fun/vba/Vba.java
+++ b/src/main/mondrian/olap/fun/vba/Vba.java
@@ -1,12 +1,11 @@
 /*
-* This software is subject to the terms of the Eclipse Public License v1.0
-* Agreement, available at the following URL:
-* http://www.eclipse.org/legal/epl-v10.html.
-* You must accept the terms of that agreement to use this software.
-*
-* Copyright (c) 2002-2013 Pentaho Corporation..  All rights reserved.
+// This software is subject to the terms of the Eclipse Public License v1.0
+// Agreement, available at the following URL:
+// http://www.eclipse.org/legal/epl-v10.html.
+// You must accept the terms of that agreement to use this software.
+//
+// Copyright (c) 2002-2015 Pentaho Corporation..  All rights reserved.
 */
-
 package mondrian.olap.fun.vba;
 
 import mondrian.olap.InvalidArgumentException;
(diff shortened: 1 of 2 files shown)


mkambol:
@akhayrutdinov 
Sorry for not catching this earlier, but there's an issue with this change, both in master and in your SP PR.  'y' and 'd' cannot be treated as the same thing in all contexts.  DatePart() should still return the DAY_OF_MONTH.  E.g. with the following two calc members:

Member Measures.[datepart-y-dec1-2001] as DatePart('y',CDate('Dec 1, 2001'))
Member Measures.[datepart-d-dec1-2001] as DatePart('d',CDate('Dec 1, 2001'))

The first should return 335, the second should return 1, and that was the behavior prior to your change.  The DatePart() tests are inadequate--I don't see a test of "d" anywhere.

SSAS and Vba must assume that within a DateDiff the user *intends* to get a raw difference of dates.

akhayrutdinov:
@mkambol, better late than never :)
What do you think - is it acceptable to make a manual assignment here: https://github.com/pentaho/mondrian/blob/master/src/main/mondrian/olap/fun/vba/Vba.java#L427
If Interval is 'd', then simply assign is to 'y'.

At first glance, after Interval.d will have been reverted, backward-compativility will be not broken and at the same time the reported issue will be solved.

akhayrutdinov:
This is what I mean: https://github.com/pentaho/mondrian/pull/492

mkambol:
@akhayrutdinov I think that seems reasonable.  

Also, can you scan down the list of other VBA functions implemented in Mondrian to see if any others seem like they might have similar, nuanced behavior?  I can run tests on SSAS if you think any are worth checking.  Don't spend a lot of time, but better to try to flush them out now then wait for the bug report.

akhayrutdinov:
@mkambol, ok, i'll take care of it. If necessary, I'll update #492

akhayrutdinov:
@mkambol, I've added changes from #492 to this PR. Review it as well please.

PR #492, opened 2015-06-23

akhayrutdinov:
- revert changes related to Interval.d from commit 1ee115751dfb9e5520d411a0b37232229728b409
- change the behaviour of dateDiff function, by manually replacing Interval.d with Interval.y
- add a test to verbose the difference between 'd' and 'y' for DatePart function (https://msdn.microsoft.com/en-us/library/20ee97hz(v=vs.90).aspx)

@mkambol, review it please. I hope this will solve the problem.
/cc @lucboudreau

### DIFF
--- a/doc/vba_functions.html
+++ b/doc/vba_functions.html
@@ -500,7 +500,7 @@ functions implemented, is described in the <a href="mdx.html">MDX specification<
 	q Quarter <br>
 	m Month <br>
 	y Day of year <br>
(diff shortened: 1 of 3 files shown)


akhayrutdinov:
@mkambol, I've updated the PR with tests representing cases with ```dateAdd``` function you ran. They are passed, so I suppose the function is not affected by the changes made for ```dateDiff```.

Review the PR please

PR #494, opened 2015-06-24
[SP-1986][MONDRIAN-2352] Mondrian incorrectly flags some NonEmptyCros…

mkambol:
…sjoins as

not supporting native evaluation.

This change refines the logic which looks for "conflicting" members in
measure expressions.  If any member in the crossjoin overlaps with a
member in a measure expression then we now allow native evaluation to
proceeed.  For example, if a measure has a nested reference to
[Time].[April], and the crossjoin contains {[March], [April]}, then we
know that native CJ involving that set won't exclude [April].

This change also removes the Head and Tail functions from the blacklist.
The members returned by Head/Tail should be a subset of those in the
argument, so should be allowed.
(cherry picked from commit bfe0e60)


mkambol:
@lucboudreau

PR #498, opened 2015-06-29
MONDRIAN-2366 - CacheControl.flush(MemberSet) does not flush NativeSet member caches.

YuryBY:
@mkambol, @lucboudreau, take a look, please

### DIFF
--- a/src/main/mondrian/rolap/CacheControlImpl.java
+++ b/src/main/mondrian/rolap/CacheControlImpl.java
@@ -1,12 +1,11 @@
 /*
-* This software is subject to the terms of the Eclipse Public License v1.0
-* Agreement, available at the following URL:
-* http://www.eclipse.org/legal/epl-v10.html.
-* You must accept the terms of that agreement to use this software.
-*
-* Copyright (c) 2002-2013 Pentaho Corporation..  All rights reserved.
+// This software is subject to the terms of the Eclipse Public License v1.0
+// Agreement, available at the following URL:
+// http://www.eclipse.org/legal/epl-v10.html.
+// You must accept the terms of that agreement to use this software.
+//
+// Copyright (c) 2002-2015 Pentaho Corporation..  All rights reserved.
 */
-
 package mondrian.rolap;
 
 import mondrian.olap.*;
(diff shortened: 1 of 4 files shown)


lucboudreau:
@mkambol Please review this PR. From a quick review, I can already tell it'll need to be reworked a bit since none of the changes are thread-safe.

mkambol:
@lucboudreau @YuryBY 
The thing that caught my eye is the implementation of .flushCache, which checks for cache size before clearing.  There's no need to check size, so that could just be a call to .clear() to avoid needing to lock.
The cache itself uses a ReentrantReadWriteLock for all operations, so I think the call to .clear() itself requires no extra handing.
Anything else you were seeing, Luc?

lucboudreau:
The method flushAllNativeSetCache iterates over elements. Is this also protected by the RRWL?

mkambol:
No, but the nativeEvaluatorMap that it's iterating over is a fixed map set at construction of the registry.  For correctness I'd say that should be made an ImmutableMap.

YuryBY:
@mkambol, @lucboudreau, PR is already updated

- there was no possibility to use unmodifiablemap because of the https://github.com/YuryBY/mondrian/blob/master/src/main/mondrian/rolap/RolapNativeRegistry.java#L79 method (is supposed calling from external point). My solution is ReentrantReadWriteLock

mkambol:
@YuryBY - My guess is nobody is using the register method, but I guess better safe then sorry.

What about using a ConcurrentHashMap instead of introducing the read and write locks in each method in RolapNativeRegistry?

YuryBY:
@mkambol - places where I'm not sure about ConcurrentHashMap are iterator usages (https://github.com/pentaho/mondrian/pull/498/files#diff-64c6891e7df227db892341b4505175d0R116, https://github.com/pentaho/mondrian/pull/498/files#diff-64c6891e7df227db892341b4505175d0R93, https://github.com/pentaho/mondrian/pull/498/files#diff-64c6891e7df227db892341b4505175d0R105).  Thread-safety will work fine here but thread while iterating will not see changes in map which can be done by another one thread (by using of put or remove methods). If it's appropriate behavior I'll replace ReentrantReadWriteLock by ConcurrentHashMap. Thanks, Yury.

PR #505, opened 2015-07-14
[ANALYZER-3017]  Drillthrough in Analyzer against a Virtual cube fails

mkambol:
When Analyzer issues a DT, it includes (by default) a RETURNS clause
including all available members.  In a virtual cube this may include
some non-conforming, inapplicable members.  To address this case,
we now capture the list of inapplicable members so they can be
specially handled.  All inapplicable members result in a placeholder
field in the DT resultset with a value of NULL.

### DIFF
--- a/src/main/mondrian/olap/DrillThrough.java
+++ b/src/main/mondrian/olap/DrillThrough.java
@@ -1,16 +1,15 @@
 /*
-* This software is subject to the terms of the Eclipse Public License v1.0
-* Agreement, available at the following URL:
-* http://www.eclipse.org/legal/epl-v10.html.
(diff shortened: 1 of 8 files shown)


mkambol:
@lucboudreau

lucboudreau:
Just asking because I've never tried it myself, but how portable is it to inject NULL values in SQL queries like this? Did you get a chance to try on SQL Server, Oracle, MySQL and others?

mkambol:
I believe it's generally supported, but there are always surprises.  
I've tried MySQL, SQL Server, Postgres, and Oracle.

An alternative is to just use an empty string.  That may be safer, but not sure.

PR #511, opened 2015-07-23
[MONDRIAN-1495] Allow Aggregate levels to use more attributes from

mkambol:
hierarchy's levels so to avoid joins to the dimension tables

This change introduces support for extra level columns when using an
explicitly defined aggregate level.  These columns can be defined by
adding ordinalColumn and captionColumn attributes on an <AggLevel>, as
well as nesting <AggLevelProperties> under an <AggLevel>.  When
constructing SqlTupleReader queries to pull level metadata we'll now
check whether any "extra" columns have a mapping to the aggregate
table.  If not, the old behavior of joining agg table to dimension
table will be used as a fall back.

### DIFF
--- a/src/main/mondrian/olap/Mondrian.xml
+++ b/src/main/mondrian/olap/Mondrian.xml
@@ -1875,9 +1875,25 @@ Revision is $Id$
                 The name of the column mapping to the level name.
             </Doc>
         </Attribute>
+        <Attribute name="ordinalColumn" required="false">
+            <Doc>
+                The ordinal column for this level.
+            </Doc>
+        </Attribute>
+        <Attribute name="captionColumn" required="false">
+            <Doc>
+                The caption column for this level.
+            </Doc>
+        </Attribute>
         <Attribute name="name" required="true">
             <Doc>
                 The name of the Dimension Hierarchy level.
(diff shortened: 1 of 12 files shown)


mkambol:
@lucboudreau

lucboudreau:
You need to update mondrian_SW.xml as well.

mkambol:
Regarding name vs. exp:  That gets handled because $name gets treated as a property.  So it would be possible to set an extra property in the AggLevel to load that value.  It would be less confusing to the user if it were explicitly captured as a separate attribute in the AggLevel, though-- I'll go ahead and make that change.

For workbench it looks like there's more substantial changes than just updating mondrian_SW.xml.  I'll tackle that after the nameColumn addition.

lucboudreau:
I'm guessing that you're talking about this: http://jira.pentaho.com/browse/PSW-220

mkambol:
AggLevelProperty is an array, which needs special handling in SchemaExplorer with a new action defined.  The other attributes also need to be enumerated in SE.

mkambol:
@lucboudreau I've updated the PR with the nameColumn change.  I'll do the workbench UI support in a separate commit/PR.

PR #525, opened 2015-08-07
[MONDRIAN-2390] - Poor performance when accessing RolapSchemaPool from many threads

akhayrutdinov:
@mkambol, @lucboudreau, review it please (actually, I suppose the JIRA case needs to be triaged first)

@julianhyde, I will highly appreciate if you also put your remarks

### DIFF
--- a/src/main/mondrian/rolap/RolapSchema.java
+++ b/src/main/mondrian/rolap/RolapSchema.java
@@ -219,6 +219,22 @@ public class RolapSchema implements Schema {
     }
 
     /**
+     * @deprecated for tests only!
+     */
+    @Deprecated
+    RolapSchema(
+        SchemaKey key,
+        ByteString md5Bytes,
+        RolapConnection internalConnection)
+    {
+        this.id = Util.generateUuidString();
+        this.key = key;
+        this.md5Bytes = md5Bytes;
+        this.defaultRole = Util.createRootRole(this);
+        this.internalConnection = internalConnection;
+    }
+
+    /**
      * Given the name of a cell formatter class and/or a cell formatter script,
      * returns a cell formatter.
      *
--- a/src/main/mondrian/rolap/RolapSchemaPool.java
+++ b/src/main/mondrian/rolap/RolapSchemaPool.java
@@ -21,8 +21,13 @@ import org.apache.log4j.Logger;
 import java.io.IOException;
 import java.lang.ref.*;
 import java.util.*;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import javax.sql.DataSource;
+
+import static mondrian.rolap.RolapConnectionProperties.JdbcConnectionUuid;
+import static mondrian.rolap.RolapConnectionProperties.PinSchemaTimeout;
+import static mondrian.rolap.RolapConnectionProperties.UseSchemaPool;
 
 /**
  * A collection of schemas, identified by their connection properties
@@ -48,6 +53,8 @@ class RolapSchemaPool {
         mapMd5ToSchema =
             new HashMap<ByteString, ExpiringReference<RolapSchema>>();
 
+    private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
+
     private RolapSchemaPool() {
     }
 
@@ -55,7 +62,7 @@ class RolapSchemaPool {
         return INSTANCE;
     }
 
-    synchronized RolapSchema get(
+    RolapSchema get(
         final String catalogUrl,
         final String connectionKey,
         final String jdbcUser,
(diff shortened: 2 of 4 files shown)


julianhyde:
The problem, I presume, is when several clients arrive at the same time and all have the same key, and creating the value, in this case the schema, is expensive. You need to identify that the key is being populated, and have the clients wait on the value. 

And a related problem is to ensure that you are not forcing a memory barrier when clients are asking for different keys.

Guava's cache is really good at this stuff. And you don't need to mess with explicit locks when you use it. We're been talking about adding Guava for a while. Maybe this is the time.

akhayrutdinov:
@julianhyde, actually, I have no information about the keys (at least for now), but it does not matter, as the method's body is fully ```synchronized```, and that means no matter has user the pooled key or not, all except the very first are forced to wait until the nimbly thread wraps up dealing with the method.

Yep, Guava is awesome, but since it is not included in the dependencies, I haven't bear it in mind. @mkambol, do you also think it is time for the library to rise in Mondrian? Guava's caches are used in ```platform```, BTW

mkambol:
As Julian mentioned we've been talking about adding guava, and everyone has been open to it.  If you want to have another go at the fix leveraging guava and say go for it.

julianhyde:
The API to use is LoadingCache, which is a sub-class of Cache. http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/cache/LoadingCache.html

I'm not sure what version of Guava you should use -- check with the rest of the team.

mbatchelor:
Whoa ... For now, let's please stay focused on not synchronizing whole methods, and using more granular locks. I really don't want to introduce new dependencies in Mondrian at this time. I am curious though - wouldn't it be quicker/better to just use the ConcurrentHashMap and remove the locks and synchronized altogether? Ultimately, we need to get caching *out* of here, and use a pluggable cache so that we could use the CacheManager in the platform (for example).

mkambol:
Julian can correct me if I'm wrong, but the advantage of something like LoadingCache is that it would put the load of a new schema in the hands of the cache, so we don't need to worry about the potential of multiple requests for the same schema.  Even with a ConcurrentHashMap we'd still need to address that, since the schema might be loading but not yet in the CHM when a new request comes in.  I think once we do start leveraging the platform cache manager we'll want to use similar loading capabilities of jcache.

But if we're not wanting to introduce new dependencies, then the approach Andrey has taken might be a reasonable alternative.

julianhyde:
If you move to a pluggable cache you may need to make lowest-common-denominator assumptions and be more conservative about concurrency. I think most of the uses of the platform's pluggable cache are to ensure that memory is shared effectively. This is more of a pool than a cache. Using a pluggable cache makes as much sense as using a pluggable implementation of ArrayList.

People designing their own concurrent data structures scares me.

Yes @mkambol is correct. If two clients perform 'get' with the same key, the value will be instantiated once. If two clients perform 'get' with different keys they probably will not hit a memory barrier (which they definitely would if a monitor was involved) and therefore there is no contention.

mkambol:
The platform Cache Marc was referring to is basically some new infrastructure that wraps a jsr107 compliant jcache implementation and allows it to be consistently configured and exposed via osgi.  Coincidentally the first implementation we did was wrapping Guava in a jcache api, but really any jcache provider (e.g. hazelcast, ehcache, infinispan) could be used.  This is a conversation for the future, but I think if we start leveraging it, Mondrian would strictly interact with the javax.cache api.  The advantage would be to externalize all the concerns about storage, distribution, ttl, etc.  I can imagine that being able to use a distributed cache of loaded schemas could have performance advantages for large enterprise customers.  JCache also has facilities for read-through caching, which could help with the concurrent schema load problem.

akhayrutdinov:
@mbatchelor, unfortunately it is incorrect to just replace current maps with CHP, as current class' semantics implies some complicated actions to be done inside a critical section. In Java 8, CHM got ```computeIfAbsent()``` method, expecting a function to create the value. In Java 7, it is absent; I can figure out a workaround via lazy proxy that conceals all logic, but that would definitely be an clumsy solution.

Comparing two alternatives - manual improving granularity with locks and using Guava's Cache - I see Guava's advantages in giving a tested and robust code and allowing to set expiration strategy for stored values (now, each time an element is accessed, it is checked whether the value is not spoiled). The latter can help to wipe out usage of https://github.com/pentaho/mondrian/blob/master/src/main/mondrian/util/ExpiringReference.java.

On the other hand, including guava only for this case is moot. I believe it should be done coincidentally with creating an improvement request like "Replace custom code with provided by Guava (after covering with proper tests)". This class can be the very first candidate for such replacement: https://github.com/pentaho/mondrian/blob/master/src/main/mondrian/rolap/RolapConnectionPool.java

Anyway, I have no objections against both approaches. Please, do not put off the final decision and let me know it.

PR #527, opened 2015-08-10
[MONDRIAN-2366] - JUnit test. Reworked.

YuryBY:
1) Old JUnit test is removed.
2) New JUnit test is added.

@mkambol, accidentally I attached the test from jira. Here is the correct one.


mkambol:
Hey Yury- one more thing, you'll notice most tests in NativeSetEvaluationTest use formatted SQL.  We've made that a rule for new tests-- it's much easier to read when you're looking at failure output.

Just set GenerateFormattedSql:
  propSaver.set(propSaver.properties.GenerateFormattedSql, true);

And update your expected value.

YuryBY:
Thanks, Matt, it's really better. Already updated.

PR #529, opened 2015-08-25
[MONDRIAN-347] - Incorrect avg calculation in aggregates

akhayrutdinov:


src/main/mondrian/rolap/RolapAggregator.java
```diff
@@ -270,7 +269,7 @@ public boolean supportsFastAggregates(
      * These aggregators are NOT singletons like the above aggregators; rather,
      * each is different because of the fact count column expression.
      */
-    protected static abstract class BaseAggor extends RolapAggregator {
+    public static abstract class BaseAggor extends RolapAggregator {
```
lucboudreau: This is not necessary. Move your test in the right package and you won't need to make it public. We don't make things public just to test them like this.

lucboudreau: Aah I see. You need it in AggStar. Nevermind :)

src/main/mondrian/rolap/aggmatcher/AggStar.java
```diff
@@ -1121,6 +1128,19 @@ public void print(final PrintWriter pw, final String prefix) {
                 pw.print("): ");
                 pw.print(generateRollupString(sqlQuery));
             }
+
+            @Override
+            public String generateExprString(SqlQuery query) {
+                String exprString = super.generateExprString(query);
+                RolapAggregator rollupAggregator = getRollupAggregator();
+                if (rollupAggregator instanceof BaseAggor) {
+                    BaseAggor agg = (BaseAggor) rollupAggregator;
```
lucboudreau: This is not a good pattern. Please see if you can make this better. Do we really need the downcast? Can it ever be a different class? Shouldn't the method be part of the RolapAggregator?

akhayrutdinov: @lucboudreau, I'm trying to solve a specific case of agg-tables. BaseAggor is declared to be the class for aggregators over agg-tables (https://github.com/pentaho/mondrian/blob/master/src/main/mondrian/rolap/RolapAggregator.java#L265). I cannot figure out how to cram desired changes into ```RolapAggregator``` and even see no sense in it. Do you have any ideas how to implement the solution w/o similar casts?

akhayrutdinov:
@lucboudreau, do you still think the proposal is unacceptable?

PR #535, opened 2015-08-30
[BACKLOG-4961] Get Schema Workbench.app working without java mac comp…

pedrofvteixeira:
…at issues

	- changed jvm version from '1.6 only' to '1.7 and above'
	- ensured file mode 755 for both JavaApplicationStub and universalJavaApplicationStub
	- added universalJavaApplicationStub to .app/Contents/MacOS
	- JavaApplicationStub first calls for set-pentaho-env.sh ( to set PENTAHO_JAVA_HOME ) and then for universalJavaApplicationStub


pedrofvteixeira:
@CodeOnCoffee , @rmansoor , @mbatchelor , @dkincade please review

PR #538, opened 2015-09-01
[MONDRIAN-2400] The call to flushSchemaCache must flush only the sche…

lucboudreau:
…ma associated to the connection.

This is a sibling PR of https://github.com/pentaho/pentaho-platform/pull/2488


lucboudreau:
@mkambol Please review.

PR #540, opened 2015-09-01
[SP-2093] [MONDRIAN-2400] Makes XmlaExtra flush only the schema assoc…

lucboudreau:
…iated to the connection passed as a parameter.


lucboudreau:
Please hold on before merging this one. Will wait for a platform build to test on master first. (09-01-2015)

PR #541, opened 2015-09-03
MONDRIAN-1032 - ValidMeasure only takes into account the first hierarchy of every dimension

YuryBY:
@mkambol, @lucboudreau, would you mind if reviewing/merging


pamval:
@mkambol, October Service Pack issue, can you please review ?

PR #542, opened 2015-09-04
[MONDRIAN-2157] - TopPercent() get a NullPointerException

YuryBY:
@mkambol, @lucboudreau, take a look, please


lucboudreau:
@pamval This case isn't included in 6.0. It has no fix version. Can you figure this one out with David? I'm at a customer site all week and the cutoff is near.

pamval:
@lucboudreau this is part of the October Service Pack (you can see that in the Active Sprint).

pamval:
Assigning to Matt since Luc is busy.

mkambol:
@YuryBY I think your fix looks right.  Can you move your test into FunctionTest, though?  There are many other similar tests in that class and it's easier to keep them together.

YuryBY:
@mkambol, the changes are done

PR #543, opened 2015-09-09
[BISERVER-5087] updated the start scripts for client tools to be the …

dkincade:
…new defaults for production deployments of min:1G max:2G


dkincade:
@rmansoor @mbatchelor @pentaho-nbaker @lucboudreau please review ...

PR #552, opened 2015-09-18
[MONDRIAN-2411] Adds roles to the cache in RolapNativeSet. This bug w…

lucboudreau:
…as introduced in 2011 with commit b65ba684e1e2467933987031e7f3f05adebe0e3c, released as 3.3.0-R. Adds tests for roles and union of roles.

PR #554, opened 2015-09-18
[SP-2139] [MONDRIAN-2411] Adds roles to the cache in RolapNativeSet.

lucboudreau:


lucboudreau:
@dkincade @mbatchelor @mkambol This is the 5.4 SP port. Waiting on confirmation of tests from #552 .

PR #564, opened 2015-09-29
[BACKLOG-5370]  "Pentaho Data Services" database type not in list

mkambol:
Updating KETTLE_PLUGIN_BASE_FOLDERS to include the plugin dir
 to make sure the dataservices plugin is found and loaded.
(cherry picked from commit 9057afe)

@dkincade


mkambol:
validated against the devci workbench dist.

PR #568, opened 2015-10-01
[CLEANUP]  Removing jdk15 compilation.  jdk16+ is required.

mkambol:
@lucboudreau


lucboudreau:
Was there no changes needed, like for the excludes of compileJdk16?

PR #571, opened 2015-10-01
[MONDRIAN-2404] - UseAggregates should apply for loading all aggregat…

kurtwalker:
…es.  ReadAggregates only applies to searching for aggregates which are not explicitly defined.


kurtwalker:
@lucboudreau please review

PR #572, opened 2015-10-01
[MONDRIAN-2405] - adding timeout check during execution of Generate f…

kurtwalker:
…unction.  Correcting usage in Filter function so that it only checks on a specific interval


kurtwalker:
@lucboudreau please review

lucboudreau:
Can we also add a test for FilterFunDef?

kurtwalker:
@lucboudreau done

PR #586, opened 2015-11-03
[MONDRIAN-2403] Severe performance issue with reports containing count

SergeyTravin:
distinct measures and compound slicers

This commit includes the following changes to reduce the overhead of
distinct-count measures being evaluated in the context of a compound
slicer:
1)  Compound predicates for the slicer are constructed in the
evaluator.  This allows the slicer predicates to be constructed a
single time in most cases, rather than once per cell request.

2) Similarly, the compound predicate string representation for the slicer
is also constructed in the evaluator.

3) AggregateFunDef updated to avoid unnecessary tuple optimization
when tuplelist is supportable within an IN list.  The tuple
optimization algorithm is expensive and we should avoid it when
unnecessary.

To support 1 and 2, the logic in RolapAggregationManager for
constructing compound predicates has been extracted into a new
CompoundPredicateInfo class.  The logic for constructing predicate
strings (i.e. getCompoundPredicateStringList) was extracted from
AggregationKey and included in CompoundPredicateInfo as well.

### DIFF
--- a/src/main/mondrian/olap/fun/AggregateFunDef.java
+++ b/src/main/mondrian/olap/fun/AggregateFunDef.java
@@ -4,7 +4,7 @@
 // http://www.eclipse.org/legal/epl-v10.html.
 // You must accept the terms of that agreement to use this software.
 //
-// Copyright (c) 2002-2014 Pentaho Corporation..  All rights reserved.
+// Copyright (c) 2002-2015 Pentaho Corporation..  All rights reserved.
 */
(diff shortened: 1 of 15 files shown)


SergeyTravin:
@lucboudreau could you review the backport for https://github.com/pentaho/mondrian/pull/555

pamval:
@lucboudreau or @mkambol please review and merge this backport to 6.0

PR #587, opened 2015-11-17
[MONDRIAN-2383] - Extra joins in Mondrian generated query after selecting columns on drill-through option from Analyzer

YuryBY:
@mkambol, @lucboudreau, could you please take a look?


YuryBY:
@mkambol, PR is reworked. Thanks.

PR #595, opened 2015-11-26
[MONDRIAN-1107] - Mondrian does not call SegmentCache.getSegmentHeaders()

akhayrutdinov:
- adopt 9ab54555427f5482c56003cfd3ddff267b8086cc
- add tests

@mkambol, @lucboudreau, review it please. This is a rework over https://github.com/pentaho/mondrian/commit/9ab54555427f5482c56003cfd3ddff267b8086cc with additional tests

### DIFF
--- a/src/main/mondrian/rolap/agg/SegmentCacheManager.java
+++ b/src/main/mondrian/rolap/agg/SegmentCacheManager.java
@@ -269,6 +269,7 @@ public class SegmentCacheManager {
 
     public final SegmentCache compositeCache;
     private final SegmentCacheIndexRegistry indexRegistry;
+    private final Set<String> starFactTablesToSync;
 
     private static final Logger LOGGER =
         Logger.getLogger(AggregationManager.class);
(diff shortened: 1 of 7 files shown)


akhayrutdinov:
@mkambol, @lucboudreau, review it please.

src/main/mondrian/rolap/agg/SegmentCacheManager.java
```diff
@@ -307,6 +308,94 @@ public SegmentCacheManager(MondrianServer server) {
         }
 
         compositeCache = new CompositeSegmentCache(segmentCacheWorkers);
+
+        // syncing elements that already exist in external cache
+        // is not possible now - have to wait until the schema has been loaded
+        // thus, let's merely store headers
+        // in-memory cache will not be double-indexed,
+        // because at this point it is empty
+        starFactTablesToSync = pickStarTablesFrom(
```
lucboudreau: This won't work. There might be new cache segments by the time this is needed. Plus it's a bad practice to build-in such stateful variables.

src/main/mondrian/rolap/agg/SegmentCacheManager.java
```diff
@@ -307,6 +308,94 @@ public SegmentCacheManager(MondrianServer server) {
...
+                }
+            }
+
+            if (!headers.isEmpty()) {
+                // cache should be changed within its thread
+                List<CellCacheSegmentCreateEvent> events = execute(
+                    new Command<List<CellCacheSegmentCreateEvent>>() {
+                        @Override
+                        public Locus getLocus() {
+                            return null;
```
lucboudreau: Pretty sure that's a no-no.

lucboudreau:
@mkambol Can you please help here? This is a critical part and needs to be done right.

mkambol:
It seems to me that it would be cleaner to synchronize all headers from the external cache in one swoop when the SegmentCacheManager is created, rather than storing a list of tables and selectively synchronizing as stars are loaded.  That would eliminate the need to store a list of "tableToSync".

For the .getLocus(), returning null isn't safe.  You can get the current locus from thread local by doing a Locus.peek().

Also, if you haven't already, it would be good to actually test this with an external cache in use and make sure it works as expected.  There's a docker file in pentaho-tacklebox which will setup biserver w/ infinispan configured:  https://github.com/pentaho/pentaho-tackle-box/tree/master/docker/biserver-ee/BISERVER-INFINISPAN

lucboudreau:
@mkambol We can't load them up front, unless at least one query has hit that schema and it is loaded. We need the schema locally in order to add external segments, so his approach is OK in that regard.

mkambol:
I guess I'm not clear on why we couldn't load them up-front-- the SegmentCacheIndex doesn't hold on to a RolapStar or anything that would require the schema to be loaded.  It needs the schema checksum, but that should be retrievable from the composite cache.  At the point we retrieve anything via the index there would be an initialized schema.  

If we do need to defer until until schema load, though, it might be preferable to do so with an immutable version of tablesToSync.  Rather than removing from tablesToSync we leave the list as is.  This should only ever be hit when a star is created, so it should be unlikely that we'd attempt to synchronize a second time for the same table.  And even if we do I'd expect the overhead is usually small compared to overall schema load.

PR #596, opened 2015-12-01
[MONDRIAN-1562] - NPE in SegmentCacheIndexImpl.loadSucceeded.

DaniilMelezhik:
#### Changes
- Fixed: added null check for headerInfo.
- Added warn to notify that 'headerMap.get(oldHeader)' returned null.
- Added 'return;' if 'headerMap.get(oldHeader)' returned null.

#### Comment
This is a fix of issue [MONDRIAN-1562]. I work with Dmitry Amelin, who described the problem in our application in comments to the issue. Please look through his comment in the issue to understand what kind of problem in this commit we've fixed.

### DIFF
--- a/src/main/mondrian/rolap/cache/SegmentCacheIndexImpl.java
+++ b/src/main/mondrian/rolap/cache/SegmentCacheIndexImpl.java
@@ -240,12 +240,27 @@ public class SegmentCacheIndexImpl implements SegmentCacheIndex {
     {
         checkThread();
 
-        LOGGER.trace(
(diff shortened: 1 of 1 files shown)


lucboudreau:
This is not a fix. It is just outputting stuff to the logs. Also, since there is no test, I suspect that you have not reproduced the issue itself and was therefore not able to fix it.

Please investigate further into what the actual cause is. I think that your fix can potentially corrupt the state of the indexes. If the code calls for an update and there is nothing to update, we need to understand why.

PR #597, opened 2015-12-03
[MONDRIAN-2433] - Filtering in analyzer report for Integer, Numeric values returns UPPER() error message with Impala

akhayrutdinov:
- explicitly cast values to string when generating regex condition for Impala
- add tests

@mkambol, @lucboudreau, review it please.

I checked several DBs to find out how do they react on similar situation.
* MySql -- implicitly converts int to string
* MsSql -- Mondrian does not support generating regex conditions for it, so no problem there
* Oracle -- works fine w/o any explicit cast
* Postgres -- looks like similar problem has been reported, it was fixed in the same manner by https://github.com/pentaho/mondrian/commit/3c5d6b7b86f09fc5ab1262afdfa30c0f8d97cb2a.

To sum up, I believe my fix is quite enough for now.

### DIFF
--- a/src/main/mondrian/spi/impl/ImpalaDialect.java
+++ b/src/main/mondrian/spi/impl/ImpalaDialect.java
@@ -198,8 +198,9 @@ public class ImpalaDialect extends HiveDialect {
                     escapeMatcher.group(1),
                     escapeMatcher.group(2));
         }
-        final StringBuilder sb = new StringBuilder();
 
+        source = "cast(" + source + " as string)";
+        final StringBuilder sb = new StringBuilder();
         // Now build the string.
         if (caseSensitive) {
             sb.append(source);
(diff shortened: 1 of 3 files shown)


akhayrutdinov:
BTW, I found this issue among Impala's tickets: https://issues.cloudera.org/browse/IMPALA-1468
It is marked as fixed for 2.1. However, its fix -- https://github.com/cloudera/Impala/commit/bc3ac2452e87cb28b6a2efab3d00944b70123005 -- does not add implicit conversion, but fixes some problem with signature deduction.

With 2.1.2 it is still impossible to use `upper` with ints and booleans:
> select upper(1);
> ERROR: AnalysisException: No matching function with signature: upper(TINYINT).

> select upper(true);
> ERROR: AnalysisException: No matching function with signature: upper(BOOLEAN).

PR #598, opened 2015-12-03
[MONDRIAN-2425] - MDX will not stop at timeout for certain Analyzer r…

YuryBY:
…eports with totals/subtotals

@mkambol, would you mind of reviewing/merging?

### DIFF
--- a/src/main/mondrian/olap/fun/CrossJoinFunDef.java
+++ b/src/main/mondrian/olap/fun/CrossJoinFunDef.java
@@ -932,6 +932,9 @@ public class CrossJoinFunDef extends FunDefBase {
             // Measure and non-All Members evaluation is non-null, then
             // add it to the result List.
             final TupleCursor cursor = list.tupleCursor();
+            final int checkCancelPeriod =
+                MondrianProperties.instance().CancelPhaseInterval.get();
+            int currentIteration = 0;
             while (cursor.forward()) {
                 cursor.setContext(evaluator);
                 for (Member member : memberSet) {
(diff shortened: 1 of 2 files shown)


mkambol:
@YuryBY I don't think this will do it.  You've added a isCancelOrTimeout check within the loop, but nothing stops the loop (just the checkData).  Worse, calling isCancelOrTimeout won't actually trip the timeout flag-- it just checks whether it's already been set.  .checkCancelOrTimeout() will set the timeout state and throw a QueryTimeoutException.

We don't want to be doing this every time within a tight loop, though.  In SqlTupleReader we used the strategy of checking at a PhaseInterval.  That might work here.  But actually I think it might be preferable to just check once before we enter the cursor.forward loop.  In the cases where I've seen problems where we spend excessive time in checkData (like the ESR associated with this bug) there have been deeply nested crossjoins, so we hit nonEmptyList many times.

YuryBY:
@mkambol, PR is updated. The timeout check is still in the loop since .checkDate on each iteration could be quite expensive. Let me know if it definitely should be moved out of the loop. Thanks.

lucboudreau:
Let's not make it even slower... please use an interval.

YuryBY:
@lucboudreau, already updated. Let me know if i have incorrectly understand what did you mean. Thanks.

mkambol:
@YuryBY - using an interval is an improvement.  I don't think checking isCancelOrTimeout() is the best option, though-- using checkCancelOrTimeout() instead will check cancel/timeout and throw an appropriate exception if needed, also managing notification.

YuryBY:
@mkambol, @lucboudreau, the update is done.Thanks.

mbatchelor:
@mkambol and @YuryBY - I don't know why this was merged without any new test cases covering the behavior change.

lucboudreau:
@mbatchelor Different PR.

https://github.com/pentaho/mondrian/pull/608

lucboudreau:
Woops wrong PR.

mkambol:
@mbatchelor You're right, this really should have a test.  That was a miss on my part.
@YuryBY take a pass at adding tests for this.  It's possible you could use a strategy similar to what was done with a9107ab
Or possibly we could test the nonEmptyList method directly by running a query and grabbing an evaluator and list to use for the test.
Let me know if you get stuck and I can assist.

PR #599, opened 2015-12-05
[MONDRIAN-2225] - Using aggregations with data restrictions return error

akhayrutdinov:
- remove hierarchy restrictions if it contains no members and was initially defined with CUSTOM access
- introduce separate methods for each OlapElement to be used in granting routine
- add tests

### DIFF
--- a/src/main/mondrian/rolap/RolapSchema.java
+++ b/src/main/mondrian/rolap/RolapSchema.java
@@ -746,138 +746,182 @@ public class RolapSchema implements Schema {
     }
 
     private Role createRole(MondrianDef.Role xmlRole) {
-        final boolean ignoreInvalidMembers =
-            MondrianProperties.instance().IgnoreInvalidMembers
-                .get();
         if (xmlRole.union != null) {
-            if (xmlRole.schemaGrants != null
-                && xmlRole.schemaGrants.length > 0)
-            {
-                throw MondrianResource.instance().RoleUnionGrants.ex();
-            }
-            List<Role> roleList = new ArrayList<Role>();
-            for (MondrianDef.RoleUsage roleUsage : xmlRole.union.roleUsages) {
-                final Role role = mapNameToRole.get(roleUsage.roleName);
-                if (role == null) {
-                    throw MondrianResource.instance().UnknownRole.ex(
-                        roleUsage.roleName);
-                }
-                roleList.add(role);
-            }
-            return RoleImpl.union(roleList);
+            return createUnionRole(xmlRole);
         }
+
         RoleImpl role = new RoleImpl();
         for (MondrianDef.SchemaGrant schemaGrant : xmlRole.schemaGrants) {
-            role.grant(this, getAccess(schemaGrant.access, schemaAllowed));
-            for (MondrianDef.CubeGrant cubeGrant : schemaGrant.cubeGrants) {
-                RolapCube cube = lookupCube(cubeGrant.cube);
-                if (cube == null) {
-                    throw Util.newError(
-                        "Unknown cube '" + cubeGrant.cube + "'");
-                }
-                role.grant(cube, getAccess(cubeGrant.access, cubeAllowed));
-                final SchemaReader schemaReader = cube.getSchemaReader(null);
-                for (MondrianDef.DimensionGrant dimensionGrant
-                    : cubeGrant.dimensionGrants)
-                {
-                    Dimension dimension = (Dimension)
-                        schemaReader.lookupCompound(
-                            cube,
-                            Util.parseIdentifier(dimensionGrant.dimension),
-                            true,
-                            Category.Dimension);
-                    role.grant(
-                        dimension,
-                        getAccess(dimensionGrant.access, dimensionAllowed));
-                }
-                for (MondrianDef.HierarchyGrant hierarchyGrant
-                    : cubeGrant.hierarchyGrants)
-                {
-                    Hierarchy hierarchy = (Hierarchy)
-                        schemaReader.lookupCompound(
-                            cube,
-                            Util.parseIdentifier(hierarchyGrant.hierarchy),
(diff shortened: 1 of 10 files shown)


akhayrutdinov:

src/main/mondrian/rolap/RolapSchema.java
```diff
@@ -746,138 +746,180 @@ private NamedSet createNamedSet(MondrianDef.NamedSet xmlNamedSet) {
...
+
+        if (memberStored == 0 && hierarchyAccess == Access.CUSTOM) {
+            if (LOGGER.isTraceEnabled()) {
+                LOGGER.trace(
+                    "Rolling back grants of Hierarchy '"
+                    + hierarchy.getUniqueName()
+                    + "' to ALL, because it contains no "
+                    + "valid restricted members");
+            }
+            role.grant(hierarchy, Access.ALL, null, null, rollupPolicy);
```
mkambol: @akhayrutdinov I don't think rolling back to ALL is a good idea.  If I've setup a role that says users can only see [Store 1], and then Store 1 doesn't happen to be present due to a data issue (or whatever), I wouldn't want my users to then be able to see *all* stores.  Probably NONE makes more sense.  

@lucboudreau, do you have an opinion here?

testsrc/main/mondrian/rolap/agg/AggregationOnInvalidRoleTest.java
```diff
@@ -0,0 +1,146 @@
...
+            + "</Role>";
+
+        return TestContext.instance()
+            .create(null, cube, null, null, null, role)
+            .withRole("Test");
+    }
+
+
+    public void test_ThrowsException_WhenNonIgnoringInvalidMembers() {
+        propSaver.set(propSaver.properties.IgnoreInvalidMembers, false);
```
mkambol: This test is order dependent.  If the other test runs first then the schema will be stored in the pool and reused.  I happened to see this fail when running it myself because the other test ran first.
Flushing the cache at the start of this test might be sufficient to address that.

src/main/mondrian/rolap/RolapSchema.java
akhayrutdinov: @mkambol, I vote for ALL, because from my perspective `ignoreInvalidMember` should be interpreted literally -- it should be ignored with no additional actions.

Assume, there are two restricted members: [Valid] and [Invalid]. In such case, we simply throws the latter away as if it wouldn't exist. So, why should the behaviour change when only [Invalid] is present?

testsrc/main/mondrian/rolap/agg/AggregationOnInvalidRoleTest.java
akhayrutdinov: Thanks for the note. I'll fix it. Let's make a decision about amending hierarchy's restrictions

src/main/mondrian/rolap/RolapSchema.java
mkambol: "ignore" is a bit of a misnomer.  In most cases we treat non-existent members as the null member, we don't truly ignore.

Regardless, the scenario I gave above seems realistic and worrying.  If a role has been defined which limits access to a member or set of members, I wouldn't think we'd ever want to assume they *really* intended to grant access to everything.  It seems more likely that the dimensional data changed in some way, but we shouldn't assume that dimensional changes imply broader access rights.  Where security is involved it's best to be conservative.

akhayrutdinov:
@mkambol, I've updated the PR with changing ALL to NONE and addition of schema flush. Please review it once again.

mkambol:
@akhayrutdinov I'm seeing some test failures, e.g. testUnionRole and testBugMondrian402 in AccessControlTest.

akhayrutdinov:
@mkambol, sorry, I missed that. Now, the test passed as I've lessened the condition: hierarchy's access level is modified only if it contains some members and all that members are invalid. Also, I added one more commit with a fix making changed tests pass on my local Windows machine.

akhayrutdinov:
@mkambol, review it please.

akhayrutdinov:
@mkambol, review it please.

mkambol:
@akhayrutdinov I don't see this when running in my IDE, but this test is failing for me from an ant test run:  test_ExecutesCorrectly_WhenIgnoringInvalidMembers
Can you see whether it also fails for you?

akhayrutdinov:
@mkambol, ok let me check that

akhayrutdinov:
@mkambol, I've run `ant test` and got nothing wrong:
```
testDistinctCountOnMembersWithNonJoiningDimensionNotAtAllLevel	Success		0.004
test_ThrowsException_WhenNonIgnoringInvalidMembers	Success		0.122
test_ExecutesCorrectly_WhenIgnoringInvalidMembers	Success		0.474
testSingleJoin	Success		0.641
``` 

Do you have any conjectures why your execution failed? A tests' interdependency? I suppose `context.flushSchemaCache();` prevents it, doesn't it?

mkambol:
@akhayrutdinov I don't have a good guess at this point.  The failure I get is due to result differences, where the results look as though the fact table was used, not the agg table.  I don't see the failure if I run the test class alone, either in ant or in my IDE-- only if I run the full suite.  Could possibly be some sort of race condition.

If you can't repro or learn anything new, let me know and I'll bang on it some tomorrow.

akhayrutdinov:
@mkambol, I managed to replicate the problem. The former test is split into two, and I added fresh connection restriction as well as flushing. Now, all tests from this PR are passing on my machine. Can you please take a look at the PR?

mkambol:
I'll take a look.  Any better guess about why it was happening?

akhayrutdinov:
I updated the PR by renaming a test class with `_` in its name, because @lucboudreau does not like them for some reasons :smile:

PR #600, opened 2015-12-05
MONDRIAN-2108: unwrap proxy of olap4jconnection instead of naive cast…

rpbouman:
… (thanks to mkhludnev https://github.com/m-khl/mondrian-sandbox/commit/1af873c844613f5ab03354fe90ddc2976da43e1a)


rpbouman:
@mkambol @pmalves Hi Guys! Just a ping in case this PR wasn't noticed.

mkambol:
@rpbouman Hi Roland!  Sorry for the delay.  I had noticed it but haven't yet had a chance to try it out.  We'll need at least one test added for this-- is that something you or the original contributor could add?

PR #601, opened 2015-12-07
[MONDRIAN-2430]  The 2 arg form of TopCount can perform badly

mkambol:
The 2 arg form of TC/BC should return the tuple set in its
natural order.  Because of this, we cannot join to the fact table
when determining the set.  Native eval can still be  valuable,
however, since member constraints and limiting sql to top N rows
can reduce set evaluation time.

This change makes the isJoinRequired method of TopCountConstraint
conditional on whether an orderByExpression is present.  It
also adds logic to determine whether the constraint is valid, since
push down with the 2 arg form of TC can only be done if
a single crossjoin arg is present.

http://jira.pentaho.com/browse/MONDRIAN-2430

@lucboudreau

### DIFF
--- a/src/main/mondrian/rolap/DefaultTupleConstraint.java
+++ b/src/main/mondrian/rolap/DefaultTupleConstraint.java
@@ -5,10 +5,9 @@
 // You must accept the terms of that agreement to use this software.
 //
 // Copyright (C) 2004-2005 TONBELLER AG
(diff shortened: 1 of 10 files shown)


lucboudreau:
I'm still running some tests w/ wingman. I'll merge it tonight if everything is good.

mkambol:
@lucboudreau PR updated to address agg table handling.

PR #602, opened 2015-12-08
[MONDRIAN-1803] - Getting Wrong Results on 3.5 while correct results on 3.4 for same MDX and schema properties also

YuryBY:
@mkambol, @lucboudreau, could you please take a look?
Here is a fallback strategy. 
In order to use it mondrian.native.unsupported.alert=ERROR property should be added to mondrian.properties

### DIFF
--- a/src/main/mondrian/rolap/RolapNativeFilter.java
+++ b/src/main/mondrian/rolap/RolapNativeFilter.java
@@ -6,7 +6,7 @@
 //
 // Copyright (C) 2004-2005 TONBELLER AG
 // Copyright (C) 2005-2005 Julian Hyde
-// Copyright (C) 2005-2014 Pentaho
+// Copyright (C) 2005-2015 Pentaho
 // All Rights Reserved.
 */
 package mondrian.rolap;
@@ -130,7 +130,7 @@ public class RolapNativeFilter extends RolapNativeSet {
             return null;
         }
         if (!FilterConstraint.isValidContext(
-                evaluator, restrictMemberTypes()))
+                evaluator, true, null, restrictMemberTypes(), fun))
         {
             return null;
         }
--- a/src/main/mondrian/rolap/SqlContextConstraint.java
+++ b/src/main/mondrian/rolap/SqlContextConstraint.java
@@ -13,6 +13,7 @@ package mondrian.rolap;
 import mondrian.mdx.MemberExpr;
 import mondrian.mdx.ResolvedFunCall;
 import mondrian.olap.*;
+import mondrian.resource.MondrianResource;
 import mondrian.rolap.RestrictedMemberReader.MultiCardinalityDefaultMember;
 import mondrian.rolap.RolapHierarchy.LimitedRollupMember;
 import mondrian.rolap.aggmatcher.AggStar;
@@ -71,6 +72,17 @@ public class SqlContextConstraint
         Level [] levels,
         boolean strict)
     {
+        return isValidContext(
+            context, disallowVirtualCube, levels, strict, null);
+    }
+
+    public static boolean isValidContext(
+        Evaluator context,
+        boolean disallowVirtualCube,
+        Level [] levels,
+        boolean strict,
+        FunDef fun)
+    {
         if (context == null) {
             return false;
         }
@@ -107,11 +119,29 @@ public class SqlContextConstraint
 
         // we can not handle all calc members in slicer. Calc measure and some
         // like aggregates are exceptions
+        int maxConstraints =
+            MondrianProperties.instance().MaxConstraints.get();
         Member[] members = context.getMembers();
-        for (int i = 1; i < members.length; i++) {
-            if (members[i].isCalculated()
-                && !SqlConstraintUtils.isSupportedCalculatedMember(members[i]))
-            {
+        int calculatedMembersQuantity = 0;
+        for (int i = 0; i < members.length; i++) {
+            if (!(members[i].isCalculated())) {
+                // nothing more to check
+                continue;
+            }
+            // dealing with calculated members
+            if (!SqlConstraintUtils.isSupportedCalculatedMember(members[i])) {
+                return false;
+            }
+            List<Member> listMembers = SqlConstraintUtils
+                .expandExpressions(members[i], null, context);
+            Set<Member> uniqueMembers = new HashSet<Member>(listMembers);
+            calculatedMembersQuantity = uniqueMembers.size();
+            if (calculatedMembersQuantity > maxConstraints) {
+                String message =
+                    MondrianResource.instance()
+                        .MembersQuantityExceedsMaxConstraints
+                        .str(String.valueOf(calculatedMembersQuantity));
+                alertNonNative(context, fun, message);
                 return false;
             }
         }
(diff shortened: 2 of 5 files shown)


src/main/mondrian/olap/fun/FilterFunDef.java
```diff
@@ -114,12 +115,20 @@ public TupleIterable evaluateIterable(Evaluator evaluator) {
...
-                        nativeEvaluator.execute(ResultStyle.ITERABLE);
+                    try {
+                        result = (TupleIterable) nativeEvaluator
+                            .execute(ResultStyle.ITERABLE);
+                    } catch (NativeEvaluationInClauseLimitException e) {
+                        // "fallback" strategy supposes
+                        // mondrian.native.unsupported.alert equals "ERROR"
+                        result = makeIterable(evaluator);
+                    }
                 } else {
```
mkambol: @YuryBY I was imagining that we would determine up-front that the constraint cannot be handled natively, probably by adding a check to isValidContext.  I believe we could exceed maxConstraints for the context constraint with other native evaluators, as well, so it would be preferable to check for validity and not construct a nativeEvaluator, rather than attempting and throwing an exception.

I had left a comment in the case about how we might be able to do that check.

src/main/mondrian/rolap/SqlConstraintUtils.java
```diff
@@ -2087,14 +2089,21 @@ private static String generateSingleValueInExpr(
...
+                    MondrianProperties.instance()
+                        .AlertNativeEvaluationUnsupported;
+                // if mondrian.native.unsupported.alert equals "ERROR"
+                // than use "fallback" strategy
+                if (alertProperty.get().equals("ERROR")) {
+                    throw MondrianResource.instance()
+                        .NativeEvaluationInClauseLimit
+                        .ex(String.valueOf(maxConstraints));
+                }
             } else {
```
mkambol: You're reinventing the wheel.  See RolapUtil.alertNonNative.

testsrc/main/mondrian/test/NativeSetEvaluationTest.java
```diff
@@ -1587,6 +1587,27 @@ public void testNativeSetsCacheClearing() {
...
+    * supposes the results of native and non-native evaluations are equal.
+    *
+    * @see <a href="http://jira.pentaho.com/browse/MONDRIAN-1803">Jira issue</a>
+    */
+    public void testNativeFilterWithLargeAggSetInSlicer() {
+      // enable fallback
+      propSaver.set(
+          MondrianProperties.instance()
+              .AlertNativeEvaluationUnsupported, "ERROR");
+      final String query = "with member customers.agg as "
```
mkambol: If set to ERROR, the MDX should *fail*.  The fallback should only be used if the property is WARN.

The idea of the property is that users who need the MDX to be nativized will prefer failure over attempting to run a query which may bog down the system, so ERROR should abort.

YuryBY:
@mkambol, PR is updated let me know if i incorrectly understood you or missed something.

mkambol:
@YuryBY We still need to fail the query if unsupported.alert=ERROR.

src/main/mondrian/rolap/RolapNativeFilter.java
```diff
@@ -227,6 +228,15 @@ NativeEvaluator createEvaluator(
...
     }
+
+    private void alertNonNative(
+        RolapEvaluator evaluator, FunDef fun, String reason)
+    {
+        if (!evaluator.getQuery().shouldAlertForNonNative(fun)) {
+            return;
+        }
+        RolapUtil.alertNonNative(getClass().getSimpleName(), reason);
+    }
```
mkambol: Would be better to either hard-code the function name here, or better--just use fun.getName().  Users reviewing the logs will find that more intelligible than the classname.

testsrc/main/mondrian/rolap/NonEmptyTest.java
```diff
@@ -6133,8 +6133,6 @@ public void testMon2202SeveralFilteredHierarchiesPlusMeasureFilter() {
     }
 
     public void testMon2202AnalyzerCompoundMeasureFilterPlusTopCount() {
-        propSaver.set(
-            propSaver.properties.AlertNativeEvaluationUnsupported, "ERROR");
         assertQueryReturns(
```
mkambol: @YuryBY 
The ERROR setting in this test is intentional.  We want to verify that the query below actually does use native evaluation.
You may need to increase the maxConstraints size of this test now that we have the check.

YuryBY: @mkambol, the reason of the test failing was not in maxConstraints. It happened because of the https://github.com/YuryBY/mondrian/blob/MONDRIAN-1803-1/src/main/mondrian/rolap/SqlContextConstraint.java#L112. Earlier it returned false without exception. I reworked the fix and now logging or an exception is provided in .isValidContext.
Additionally was added one more test scenario in the test method.

src/main/mondrian/rolap/SqlContextConstraint.java
```diff
@@ -108,13 +120,29 @@ public static boolean isValidContext(
...
             }
         }
+        // see MONDRIAN-1999
+        int maxConstraints =
+            MondrianProperties.instance().MaxConstraints.get();
+        int supportedCalculatedMembersQuantity =
+            SqlConstraintUtils
+                .expandSupportedCalculatedMembers(
+                    members, context).length;
+        if (supportedCalculatedMembersQuantity > maxConstraints) {
```
mkambol: @YuryBY Not sure this count will be correct when multiple calculated members are in context.  E.g. an aggregate on Customer and another aggregate on Store.  The IN list for both attributes individually may fall below maxConstraints.

YuryBY: @mkambol, you are absolutely right. I reworked the fix in order to verify calculation members quantity for each aggregation and added appropriate tests.Thanks.

testsrc/main/mondrian/test/NativeSetEvaluationTest.java
```diff
@@ -1647,6 +1662,140 @@ public void testNativeSetsCacheClearing() {
...
+   * and no ERROR messages are occurring.
+   * After that we assign MaxConstraints to 9
+   * (less than the sum of calc members for
+   * [Store].[Store State].Members) and verify that
+   * exception was thrown and log contains an error message. Beforehand we set
+   * AlertNativeEvaluationUnsupported to ERROR.
+   * @see <a href="http://jira.pentaho.com/browse/MONDRIAN-1803">Jira issue</a>
+   */
+  public void testNativeFilterWithLargeAggSetInSlicerTwoAggs() {
+    String query =
```
mkambol: This test will fail with aggregates enabled.  
In other places we've just added a check for whether aggs are enabled at the top of the test and aborted.  Alternatively you could add the aggregate SQL

testsrc/main/mondrian/test/NativeSetEvaluationTest.java
```diff
@@ -1647,6 +1662,140 @@ public void testNativeSetsCacheClearing() {
...
+        + "    `time_by_day`.`the_year` = 1997\n"
+        + "and\n"
+        + "    `customer`.`country` in ('Canada', 'Mexico', 'USA')\n"
+        + "group by\n"
+        + "    `customer`.`gender`\n"
+        + "having\n"
+        + "    ((sum(`sales_fact_1997`.`unit_sales`) > 135000)) \n"
+        + "order by\n"
+        + "    ISNULL(`customer`.`gender`) ASC, `customer`.`gender` ASC";
+    SqlPattern oraclePattern =
```
mkambol: Point of cleanliness- can you correct the SqlPattern identifier?

src/main/mondrian/rolap/SqlContextConstraint.java
```diff
@@ -107,11 +119,28 @@ public static boolean isValidContext(
...
+            }
+            // dealing with calculated members
+            if (!SqlConstraintUtils.isSupportedCalculatedMember(members[i])) {
+                return false;
+            }
+            // http://jira.pentaho.com/browse/MONDRIAN-1999
+            int calculatedMembersQuantity =
+                SqlConstraintUtils
+                    .expandExpressions(members[i], null, context).size();
+            if (calculatedMembersQuantity > maxConstraints) {
```
mkambol: This is not quite right if the calculated member has a compound aggregate, like Aggregate( CrossJoin( set1, set2))
As it turns out native evaluation mishandles that elsewhere as well, though.   I just logged MONDRIAN-2441 for that.
Whatever we do to fix 2441 will need to update how this count is calculated as well.  I've added a comment to the case.

YuryBY: PR has been updated. Appropriate test is added. The reason why all 75 combinations of srore.names and genders don't appeared in the "in" clause is here: https://github.com/pentaho/mondrian/blob/master/src/main/mondrian/rolap/SqlConstraintUtils.java#L673 - it simply removes duplicates from the list. I've done the same here.

pamval:
@mkambol @lucboudreau can you review this? This has been apparently forgotten

mkambol:
@pamval IIRC there was at least one failed test associated with this change.  I was going to help Yury try to resolve, but haven't been able to dig into it.  I'd suggest the PR be closed until the failure(s) are resolved.

YuryBY:
@mkambol, @pamval, @lucboudreau, it possible that failed test shows another issue not related to the fix. But until we don't figure out it the PR can be closed.

PR #607, opened 2015-12-14
[MONDRIAN-2436] - SQL Injection through MDX Filter

akhayrutdinov:
- prohibit invalid decimal literals in NumberSqlCompiler
- add tests

@lucboudreau, @mkambol, @kurtwalker, review it please.

### DIFF
--- a/src/main/mondrian/rolap/RolapNativeSql.java
+++ b/src/main/mondrian/rolap/RolapNativeSql.java
@@ -12,6 +12,7 @@ package mondrian.rolap;
 
 import mondrian.mdx.*;
 import mondrian.olap.*;
+import mondrian.olap.fun.MondrianEvaluationException;
 import mondrian.olap.type.MemberType;
 import mondrian.olap.type.StringType;
 import mondrian.rolap.aggmatcher.AggStar;
@@ -130,6 +131,14 @@ public class RolapNativeSql {
             }
             Literal literal = (Literal) exp;
             String expr = String.valueOf(literal.getValue());
+            try {
+                // MONDRIAN-2436: reject everything except valid decimals
+                Double.parseDouble(expr);
+            } catch (NumberFormatException e) {
+                throw new MondrianEvaluationException(
+                    "Expected to get decimal, but got " + expr);
+            }
+
             if (dialect.getDatabaseProduct().getFamily()
                 == Dialect.DatabaseProduct.DB2)
             {
--- a/testsrc/main/mondrian/rolap/RolapNativeSqlInjectionTest.java
+++ b/testsrc/main/mondrian/rolap/RolapNativeSqlInjectionTest.java
@@ -1,0 +1,43 @@
+/*
+// This software is subject to the terms of the Eclipse Public License v1.0
+// Agreement, available at the following URL:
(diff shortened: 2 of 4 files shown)


kurtwalker:
@akhayrutdinov What is the exception when we return null?  Is it informative so the user can figure out what the problem is?  Can we assert on the exception message inside the catch of your test?

akhayrutdinov:
@kurtwalker, it that case it rejects native evaluator and fails inside non-native. I think the exception is not enough informative, but among all compilers no one throws exceptions, so I didn't do it either.

kurtwalker:
@akhayrutdinov looks like the message is "Expected value of type STRING; got value '0.0' (NUMERIC)"  Can you figure out what it's not "Expected value of type STRING; got value 'null' (NULL)" or something like that.  A quick look through the code and that's what I would expect.

akhayrutdinov:
@kurtwalker, that's expectable. `(select 100)` is classified as String, hence there is nothing surprising Mondrian tries to cast read values to it. It reads tuples in non-native evaluator. "0.0" because a member has that value as selected measure

akhayrutdinov:
@kurtwalker,

>  I think the exception is not enough informative, but among all compilers no one throws exceptions, so I didn't do it either. 

should be read as **is enough** :) So, if it is ok to throw exceptions inside compilers, I'd better do it

kurtwalker:
@akhayrutdinov since we are returning a null from the compile, I don't understand how the object received at the point of exception is a Numeric instead of a null.

akhayrutdinov:
@kurtwalker, non-native evaluator tries to read all affected members and then it is comparing them with filter condition. Returned `null` means rejection of compilation to native sql, but not the read-from-db value.

kurtwalker:
@akhayrutdinov in GenericCalc line 91, the object returned from evaluate is apparently of type Number.  I'd like to understand where it got changed from null to 0.0 and if there is anything we could do about that.

akhayrutdinov:
@kurtwalker, amazingly that is `RolapUtil.valueNotReadyException` returned from `FastBatchingCellReader.get(RolapEvaluator)`. Looking at this, I do think throwing exception in native compiler makes sense as we claim the query is not valid. Do you agree? Is there any appropriate exceptions? May be `MondrianEvaluationException` fits here?

kurtwalker:
@akhayrutdinov Great.  Yes, I think MondrianEvaluationException is the correct exception to use.

lucboudreau:
@akhayrutdinov The valueNotReadyException means that the cell value you've tried to compute is not in cache, but will be fetched on the next phase. Most functions simply continue to work as usual when they encounter this value. Some of them will trigger a phase manually in order to get the right value right away because they cannot continue without it. You should look at other usages of valueNotReadyException to determine how to play nicely with it.

akhayrutdinov:
@lucboudreau, thanks for the explanation!
@kurtwalker, I've updated the PR

kurtwalker:
@akhayrutdinov can you add an assertEquals for the exception message in your test?  Want to avoid a false positive if we can some other exception

akhayrutdinov:
@kurtwalker, done

akhayrutdinov:
@kurtwalker, done

PR #618, opened 2015-12-21
[BACKLOG-5898] - Adding an "Internal Use" annotation to the implicit "Fact Count" measure.

rfellows:
This will enable a better way for determining which "visible=false" elements should always remain hidden from an Analyzer user. We can look for that annotation rather than a specific named element (like Fact count).


rfellows:
@kurtwalker please review

PR #635, opened 2016-01-25
Mondrian 2438

ivanpogodin:
@mkambol , could you review it, please?
I have a doubt about the line
https://github.com/pentaho/mondrian/compare/master...ivanpogodin:MONDRIAN-2438?expand=1#diff-1c1415254ae1023ca8f35b5f35499a21R369
This condition is made required but it might be checked instead.

### DIFF
--- a/src/main/mondrian/rolap/SqlConstraintUtils.java
+++ b/src/main/mondrian/rolap/SqlConstraintUtils.java
@@ -286,7 +286,7 @@ public class SqlConstraintUtils {
     /**
      * Creates a predicate for the slicer tuple list
      */
-    private static StarPredicate getSlicerTuplesPredicate(
+    static StarPredicate getSlicerTuplesPredicate(
         TupleList tupleList,
         RolapCube baseCube,
         AggStar aggStar,
--- a/src/main/mondrian/rolap/SqlConstraintUtils.java
+++ b/src/main/mondrian/rolap/SqlConstraintUtils.java
@@ -757,7 +757,7 @@ public class SqlConstraintUtils {
         return members;
     }
 
-    private static Member replaceCompoundSlicerPlaceholder(
+    static Member replaceCompoundSlicerPlaceholder(
         Member member,
         RolapEvaluator evaluator)
     {
--- a/src/main/mondrian/rolap/SqlConstraintUtils.java
+++ b/src/main/mondrian/rolap/SqlConstraintUtils.java
@@ -6,7 +6,7 @@
 //
 // Copyright (C) 2004-2005 TONBELLER AG
 // Copyright (C) 2005-2005 Julian Hyde
-// Copyright (C) 2005-2015 Pentaho and others
+// Copyright (C) 2005-2016 Pentaho and others
 // All Rights Reserved.
 */
 package mondrian.rolap;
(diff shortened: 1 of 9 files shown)


ivanpogodin:
Newly created QueryException is thrown.
some renames are applied.

No i18n in error messages, since it's not used in the existing code.
I'm in trouble, how to test throwing the exception.

ivanpogodin:
@mkambol , could you review it again, please?

ivanpogodin:
@mkambol , could you review it , please
Exception test is added.
I'll squash if everything is correct

mkambol:
@lucboudreau would be good to have your eyes on this one, too.

lucboudreau:
Exceptions should be defined in MondrianResource.xml and not in their own class. We generate the exception classes automatically. If you want to i18n them, you can add these in MondrianResource_[lang].xml. See example below:

```
<exception id="601010" name="InvalidHierarchyCondition" className="mondrian.olap.InvalidHierarchyException">
   <text>Hierarchy ''{0}'' is invalid (has no members)</text>
</exception>
```

lucboudreau:
I'm seeing two test failures with this PR.

ivanpogodin:
@mkambol , @lucboudreau , could you review it , please
Added SqlConstraintUtils.expandSupportedCalculatedMember(Member, Evaluator)
Exception implemented via MondrianResource.xml

testOrderCalc and testTuple were failing before this PR as well. I'm investigating.

mkambol:
testOrderCalc and testTuple are not failing for me on master, or on http://ci.pentaho.com/job/mondrian/.  Those will need to be resolved before merging.

src/main/mondrian/rolap/SqlConstraintUtils.java
```diff
@@ -757,6 +757,22 @@ public static boolean hasMultiPositionSlicer(List<Member> slicerMembers) {
         return members;
     }
```
mkambol: The new expandSupportedCalculatedMember is basically the same as the block inside the loop of the original.  Would be nice to eliminate duplication.

One thing the new method omitted was handing disjoint tuples, however.  That may be something we need to disallow here as well-- not completely clear to me.  Have you tested that condition?

Also, this new method has an issue that bugged me with the original-- if a calculated member is unexpandable it will just return that calculated member.  So you may get back a list of base members, but you also might get back a calc member or empty list.  I wouldn't be surprised if there are some existing bugs lurking because that wasn't expected.  We should at least have some clear javadoc that says what this method will return given the various conditions.  

It seems like there should be a nice simple way to ask for a clean list of fully expanded base members, if possible, with some way of knowing that expansion was not fully possible if it fails.

ivanpogodin:
@mkambol , @lucboudreau
These tests are irregular
Depend on names of anonimous classes
http://jira.pentaho.com/browse/MONDRIAN-2454

mkambol:
@ivanpogodin Nice find.  We should probably make the validation check for compiled expressions ignore anonymous class naming.

ivanpogodin:
@mkambol , @lucboudreau , could you review it again, please
Tests are corrected: https://github.com/pentaho/mondrian/pull/645 .
Commit https://github.com/ivanpogodin/mondrian/commit/9a8b957777c91258424dd35270f3897f6f9c7fd5 is shared with it

ivanpogodin:
@mkambol , @lucboudreau , 
it's rebased to the master head.

src/main/mondrian/rolap/SqlConstraintUtils.java
mkambol: Hey @ivanpogodin, I'd still like to see the above addressed before processing the PR.

ivanpogodin: @mkambol , ok

ivanpogodin:
@mkambol , could you review it?
duplicated code is refactored
tests are added beforehand to be sure, the behavior is kept unchanged

ivanpogodin:
@mkambol , I forgot one point:
not registered new test in Main

ivanpogodin:
@mkambol , 
>>not registered new test in Main
done now

mkambol:
@ivanpogodin This still doesn't address disjoint or multi-position calculations in compound slicers.  E.g. the query below will give incorrect results.  We need to either explicitly forbid such scenarios, or handle them.

```
with member time.agg as 'Aggregate( { ( Gender.F, Time.[1997].Q1), (Gender.M, Time.[1997].Q2) })
select measures.[customer count] on 0 
from sales 
where {time.agg, Time.[1998]}
```

Also, FunctionTest is probably not the best place for your tests.  CompoundSlicerTest is probably more appropriate, or add a new class.

ivanpogodin:
@mkambol , could you review it?
A new test is added to CompoundSlicerTest but previous tests weren't moved  there from FunctionTest.

PR #651, opened 2016-02-10
[MONDRIAN-2098] Alias error received when using an SQL Expression in …

SergeyTravin:
…Mondrian Schema against a Greenplum database


SergeyTravin:
@lucboudreau, @mkambol, could you review  it please?

mkambol:
@SergeyTravin Did you run the test suite against greenplum with this change?  There will likely be errors, but would be good to run before and after to verify this doesn't introduce any new issues with particular SQL forms.

SergeyTravin:
@mkambol, could you review it again?  the some test was fixed.

PR #654, opened 2016-02-11
[TEST] - forcing operation inside filter to take a long time so that …

kurtwalker:
…the timeout will always occur inside the filter loop and we always get the same exception


kurtwalker:
@mkambol @lucboudreau please review

PR #655, opened 2016-02-12
[CLEANUP] - we do not have log4j.properties checked in.  using log4j.…

kurtwalker:
…xml for default in cmdrunner instead


kurtwalker:

PR #660, opened 2016-02-16
[MONDRIAN-2308] - covering additional code path for parent-child leve…

kurtwalker:
…l dot members


kurtwalker:
@lucboudreau if we decide to merge for 6.1 this is ready

kurtwalker:
@mkambol I think @lucboudreau might be traveling today.  So, if you agree this is low risk can you merge?  thanks.

PR #661, opened 2016-02-19
[MONDRIAN-2451] Mondrian generates ORDER BY sql that is invalid with default settings of MySQL 5.7.5+

mr9d:
- Ability to generate GROUP BY expressions for nullable fields is added to Dialect interface and SqlQuery class
- UTs fixed

### DIFF
--- a/src/main/mondrian/rolap/SqlMemberSource.java
+++ b/src/main/mondrian/rolap/SqlMemberSource.java
@@ -616,7 +616,7 @@ RME is this right
         hierarchy.addToFrom(sqlQuery, level.getKeyExp());
         String q = level.getKeyExp().getExpression(sqlQuery);
         String idAlias =
-            sqlQuery.addSelectGroupBy(q, level.getInternalType());
+            sqlQuery.addSelectGroupBy(q, level.getInternalType(), true);
 
         // in non empty mode the level table must be joined to the fact
         // table
--- a/src/main/mondrian/rolap/sql/SqlQuery.java
+++ b/src/main/mondrian/rolap/sql/SqlQuery.java
@@ -484,6 +484,7 @@ public class SqlQuery {
      * the GROUP BY clause, if the dialect requires it.
      *
      * @param expression Expression
+     * @param type Java type to be used to hold cursor column
      * @return Alias of expression
      */
     public String addSelectGroupBy(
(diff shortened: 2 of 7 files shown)


mr9d:
@mkambol please review.

mkambol:
@mr9d Did you try changing the requiresOrderByAlias() for the MySQL dialect?

It worries me to be making a change to the GROUP BY structure for *all* database types.  It will be very difficult to verify this doesn't cause issues with some of them (I'd expect this will break Pentaho Data Service datasources, for example).

mr9d:
@mkambol I've tried simply override requiresOrderByAlias() in the MySqlDialect and make in always return true.

That solves the issue with TestAggregationManager.testLevelKeyAsSqlExpWithAgg, but causes an amount of new failures: some of them is simply expected SQL statements, but some is more complicated:

* mondrian.rolap.VirtualCubeTest.testNonEmptyCJConstraintOnVirtualCube
* mondrian.rolap.VirtualCubeTest.testNonEmptyConstraintOnVirtualCubeWithCalcMeasure
* mondrian.test.CompoundSlicerTest.testBugMondrian899
* mondrian.test.NamedSetTest.testHierarchizeNamedSetImmutable
* mondrian.test.NamedSetTest.testNamedSetRangeInSlicerPrimed
* mondrian.rolap.SharedDimensionTest.testVirtualCube
* mondrian.rolap.NonEmptyTest.testVirtualCubeMembers
* mondrian.rolap.NonEmptyTest.testVirtualCubeCrossJoin
* mondrian.rolap.NonEmptyTest.testVirtualCubeNonEmptyCrossJoin
* mondrian.rolap.NonEmptyTest.testVirtualCubeCrossJoinCalculatedMember2
* mondrian.rolap.NonEmptyTest.testVirtualCubeCrossJoinCalculatedMember1

Failures details: https://drive.google.com/file/d/0B7nxKOjLPqMiNWFORklvQl9fUU0/view?usp=sharing

For now I tend to stick with my initial solution which doesn't requires a lot of changes in unit tests.

lucboudreau:
I'm a little worried about this change. Why do we need to group nulls differently when this case is about the order clause? Using CASE statements also makes SQL queries much slower and hard to optimize; it's always preferable to use ANSI null ordering when available. 

Overall, this PR seems to outreach and touch things it should not.

mkambol:
@lucboudreau I'd agree with that.  My earlier comment pointed out that this change will break Data Services datasources, and may break other datasources that don't happen to support such an expression in the group by (or don't support as we expect).

I'd fully expect the more narrow fix of using the order by alias for MySql to break lots of tests, but most should be simply because the SQL has changed.  I haven't looked closely at the "more complicated" failed tests you listed yet, but it looks like we have a problem constructing the UNION'd SQL for virtual cubes (I think that's in SqlTupleReader).  That may be a problem that happens with other RDBMs that require an alias for the column, so this could indicate a more general problem.  Can you investigate those failures and get back to us?

mr9d:
@mkambol sure, I will take a close look on them tomorrow.

This pull request can be closed, I suppose.

PR #668, opened 2016-02-29
[MONDRIAN-2468] MemberFormatter not applied using member.getCaption

alostale:
In case a member defines a formatter apply it when getting a localized
  caption.


lucboudreau:
No test? :)

alostale:
Hi,

Sorry, but I'm not familiar with Mondrian infrastructure :( So I couldn't create the test case.

I reproduced the issue with Saiku + Mondrian http://jira.meteorite.bi/browse/SKU-1439 and this patch solved the problem for me though I'm not 100% it's correct.

PR #669, opened 2016-03-03
[MONDRIAN-2451] Mondrian generates ORDER BY sql that is invalid with default settings of MySQL 5.7.5+

mr9d:
This is the second attempt to resolve MONDRIAN-2451. Previous pull request was rejected: https://github.com/pentaho/mondrian/pull/661

Accordint to Matt Campbell's suggestion, I've added requiresOrderByAlias()=true for the MySQL dialect. This caused a side effect with "ORDER BY ISNULL(1) ASC, 1 ASC, ISNULL(2) ASC, 2 ASC, etc" statements in SqlTupleReader.java so it was needed to change them to "ORDER BY ISNULL('c0') ASC, 'c0' ASC, ISNULL('c1') ASC, 'c1' ASC, etc".

Still not sure about this change, because of Luc Boudreau's comment in source code:

> Sort the union of the cubes.
> The order by columns need to be numbers,
> not column name strings or expressions.

(SqlTupleReader.java:838)

Anyway, all tests are successful now, so it might be the an answer to MONDRIAN-2451.


mr9d:
@mkambol @lucboudreau please review.

src/main/mondrian/rolap/SqlTupleReader.java
```diff
@@ -841,7 +841,7 @@ private void savePartialResult(List<List<RolapMember>> partialResult) {
                 if (fullyJoiningBaseCubes.size() > 1) {
                     for (int i = 0; i < types.size(); i++) {
                         unionQuery.addOrderBy(
-                            i + 1 + "",
+                            "c" + i,
                             true,
```
mkambol: We probably want to make this change conditional on whether the dialect has requiresOrderByAlias=true.  It's possible (likely) that some databases which don't require aliases will fail with the new UNION structure if we don't have a numbered order by list.

And to confirm-- you've tried older versions of MySQL with this change to the UNION sql?  (and full test suite?)

mkambol:
@lucboudreau, do you remember the background of your comment the "order by columns need to be numbers"?  Are there specific databases where this is true?

mkambol:
@mr9d  I think we need some new tests that explicitly cover the scenario we're fixing.  testLevelKeyAsSqlExpWithAgg fails with the ORDER BY containing an expression, but that was sort of an accidental catch.   

It would be good to have more test cases that uses a sql expression in a Level definition, and then explicitly hit the various ways an ORDER BY could occur-- including with a virtual cube.

src/main/mondrian/rolap/SqlTupleReader.java
mr9d: Yep, I've tried full test suite locally on MySQL 5.7.10 and on MySQL 5.5 on CI-server.

lucboudreau:
@mkambol Some DBs (Oracle?) need the parent query to order by column ordinals rather than their alias.

mkambol:
Thanks @lucboudreau, that's what I was guessing.

@mr9d - Looking at this again, why is it we need the UNION'd order by to use aliases?  Using the ordinal should still work just fine.  The reason those tests were failing after the dialect change was because we started wrapping the ordinal in quotes.  If we just make sure the UNION'd order by SQL is still valid then it seems like it should be fine to continue using the ordinal.

mr9d:
@mkambol I've had the same idea, but was unable to test it on CI because of server unavailability. Now I see that it might be an another solution.

Please review the latest changes: I've reverted SqlTupleReader back to ordinals usage for UNION queries and updated SqlQuery to make sure that addOrderBy() method will use aliases only if alias is not null and it will be possible to use expressions even if requiresOrderByAlias() is true.

mkambol:
This is looking better, @mr9d.  Can you do two things?
1)  Add test cases that use a sql expression in a Level definition, and then explicitly hit the various ways an ORDER BY could occur-- including with a virtual cube.
2)  Add a comment to the .requiresOrderByAlias method in MySqlDialect that references the jira case and describes the scenario where an alias is actually required-- i.e. ISNULL wrapping a SQL expression in an ORDER BY can fail.  If I were reviewing the dialect without knowing that bit of trivia I would think the method incorrect.

mkambol:
Also, please squash all these commits.

mr9d:
@mkambol please review. Also I've sent some details via e-mail.

mr9d:
@mkambol please review.

mkambol:
@mr9d We're getting closer.  Try running the test suite with aggregate tables turned on.  There's a few failures.

mondrian.rolap.aggregates.Use=true
mondrian.rolap.aggregates.Read=true

mr9d:
This is done.

mr9d:
@mkambol could you please merge?

PR #670, opened 2016-03-07
[CLEANUP] Adds the bundle plugin to the maven build script so that we…

lucboudreau:
… produce valid OSGI bundles through the maven build.


lucboudreau:
@kurtwalker @mkambol For testing please. I've built it w/ Maven, dropped it into a BA server and everything seems to still work correctly.

mkambol:
Looks good, headers appear correct and works from baserver for me, too.

PR #673, opened 2016-03-15
[MONDRIAN-2443] Properties of type float not formatted properly from …

mchen-len-son:
…a SQL Server DB

- fixed rolap member property formatting
- added unit test

### DIFF
--- a/src/main/mondrian/rolap/RolapMemberBase.java
+++ b/src/main/mondrian/rolap/RolapMemberBase.java
@@ -5,7 +5,7 @@
 // You must accept the terms of that agreement to use this software.
 //
 // Copyright (C) 2001-2005 Julian Hyde
-// Copyright (C) 2005-2013 Pentaho and others
+// Copyright (C) 2005-2016 Pentaho and others
 // All Rights Reserved.
 */
 package mondrian.rolap;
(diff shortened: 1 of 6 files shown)


mchen-len-son:
@lucboudreau, @mkambol, @pamval 

Guys, please, take a look a the changes.

I see the previous formatting implementation was locale-insensitive due to _BigDecimal.toString_ javadoc. So I left this behavior unchanged.
But, is this correct? Maybe we should make it locale-sensitive?

mkambol:
One thing I'd hope to see is that a numeric column used as a member caption should have the same formatting as if that column were used instead as a member property-- we should be consistent.  I could be wrong, but I think numeric member captions would not get wrapped by BigDecimal in the same way and the formatting could be different.  Is that correct, @lucboudreau?

mchen-len-son:
@lucboudreau 

Did you have a chance to take a look at the PR?

lucboudreau:
Had to do a bit of research. I was a bit worried about using a String representation with ``val.toString()``, but my previous code was even worse, and is introducing loss of precision.

Anyhow, to Matt's comment, it would be nice to apply this logic to member captions as well. Although IIRC the code related to members uses a default implementation of ``MemberFormatter``, unlike here, where the code is procedurally included when there are no formatters for the current property. 

If we're doing a true reconciliation of these, let's make sure they also share the same code. Refactor the default formatting of captions and properties into one method if possible (you will need to do a bit of research, there might be some different rules making this whole thing more work than required)

mchen-len-son:
@mkambol, @lucboudreau, I've updated the changes. Review, please.
- added formatting for caption column numeric values

mkambol:
Hey @mchen-len-son - looks good.

One thing, though, you'll need to add your new test class to mondrian.test.Main for it to be included in the test suite.

mchen-len-son:
@mkambol, thanks. 
I've added test to the test suite.

PR #674, opened 2016-03-23
[MONDRIAN-2467] ValidMeasure with a null tuple throws NPE

ivanpogodin:


ivanpogodin:
@mkambol , could you review it, please?

By the way
If there is no Measure in ValidMeasure arguments, the error is NPE. It's better to throw a more helpful exception, isn't it?

ivanpogodin:
@mkambol , @kurtwalker , @lucboudreau , could you review it?
The customer has requested an escalation on this JIRA.
http://jira.pentaho.com/browse/MONDRIAN-2467

kurtwalker:
Did you experiment with returning an empty list instead of null?  May be inviting a different null pointer exception later by returning null here

ivanpogodin:
@kurtwalker ,
Yes, I've checked it.
getCalcsMembers is a private method, there is the null control after calling it
https://github.com/pentaho/mondrian/pull/674/files#diff-65efb9fa344d708455af9d59b6928008R72

kurtwalker:
great.  :+1:

mkambol:
@ivanpogodin, regarding having no measure, I would actually expect the default measure to be used in that scenario, since a tuple should implicitly include the default member of each hierarchy not explicitly referenced.  SSAS returns a NULL in that case, though, so might be better to mimic their behavior.

Either way, throwing a NPE or OOB is not great.  Can you log a bug?

PR #675, opened 2016-03-23
BACKLOG-5895 -- Expose Relational Information through Mondrian

kolinus:
Utility/Helper

What was done:
1) exposed xml description of RolapCubeDimension
2) added utility

### DIFF
--- a/src/main/mondrian/olap4j/DBInfoOlap4jHelper.java
+++ b/src/main/mondrian/olap4j/DBInfoOlap4jHelper.java
@@ -1,0 +1,149 @@
+/*
+// This software is subject to the terms of the Eclipse Public License v1.0
+// Agreement, available at the following URL:
+// http://www.eclipse.org/legal/epl-v10.html.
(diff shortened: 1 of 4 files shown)


kolinus:
@kurtwalker @rfellows  could you please take a look

kolinus:
@kurtwalker @lucboudreau  could you please review.

lucboudreau:
Since the methods in ``DBInfoOlap4jHelper.java`` are all static, you might as well move the whole class in the ``mondrian.rolap`` package, then revert your scope changes to the other classes.

Secondly, why does this need to be in mondrian? Whatever uses this should have the class within it. It should work as long as it's in the same package, ``mondrian.rolap``.

kolinus:
@lucboudreau Luc, currently we need this for analyzer where we have no mondrian packages.
Also this stuff is useless without mondrian, so it looks logical to be placed to mondrian.

lucboudreau:
Please move this to Analyzer and create the missing package.

PR #681, opened 2016-04-01
[MONDRIAN-2443] Properties of type float not formatted properly from …

mchen-len-son:
…a SQL Server DB

- fixed rolap member name formatting
- got rid of DecimalFormat in number formatter
- rearranged number formatter factory, hid the default implementation
- added/fixed unit tests


mchen-len-son:
@mkambol @lucboudreau 

This is an additional PR.
It appeared, that not only property and caption column but the level itself is not being properly formatted. 
Considering that rolap member name is requested much more often during report generating, I got rid of DecimalFormat as it known as performance affecting one. 

Please, review the changes.

eliofreitas:
Rejected from automatic build ,authorized users only

lucboudreau:
@mkambol Matt, please do a throughout review, as you've been on this one. My two cents: Let's not call this a number formatter. We use more generic formatters, like MemberFormatter or PropertyFormatter, which are not tied to a particular type. Also, adding a constructor to RolapMemberBase is making me itchy. Probably not the best architectural solution.

julianhyde:
> Also, adding a constructor to RolapMemberBase is making me itchy. Probably not the best architectural solution.

I agree. Especially if this would be used for a hierarchy that has lots of members. Creating a member needs to be quick. And also, members need to be small.

mchen-len-son:
Hey guys,
agreed, we need to inscribe the changes in a more suitable way.

@lucboudreau you're saying:
>  the formatters should be part of the schema

Did you mean that we should create separate formatters for the each case? As you can see by the changes, we're adding formatting for caption column values and for level values (key formatting in RolapMemberBase). 

I.e., we want to properly format:
- **property values**
- **captions**
- caption column values
- values (keys)

For the two first items we already have corresponding formatters, for others - don't.

As I see, we could create missing formatters (add them to Schema as well), and then implement a default formatter for each interface, which should be used if not specified a custom one. Thus, we would stay consistent with formatting rules for every member property and inscribe the changes into the existing code.

+@mkambol 
What do you think, guys?

mchen-len-son:
@lucboudreau @mkambol 
guys, did you have a chance to take a look at the message?
does that make sense for you?

lucboudreau:
> we could create missing formatters (add them to Schema as well), and then implement a default formatter for each interface, which should be used if not specified a custom one

That's exactly what I meant. We need a default implementation that does nothing (or the minimum), and let the users the possibility to plug-in their own formatter through the schema, like we do with MemberFormatter.

eliofreitas:
Hello, This is an automated comment from our build system. Since this pull request originated from a user  not in the Pentaho organization, we will not proceed  with automatic validation of the pull request.  However, rest assured that it will be reviewed manually as soon  as we get the chance. Thanks for your contribution!

mchen-len-son:
updated the PR:
- added **Member Key Formatter**
- added **Member Caption Column Formatter**
- added Default Formatter implementation which is in fact a NumberFormatter
- added Formatter Factory
- moved formatter creating to a single place in factory
- **Member key formatter**, **Caption column formatter** and **Member properties formatter** have a default implementation from now on in case no custom formatter is specified
- refactored related code a bit: formatter factory is instantiated once when RolapCube is initializing
- added unit tests

@mkambol @lucboudreau please, review the changes

lucboudreau:
@mkambol We'll need to review this together. I see a lot of potential for breaking stuff here.

It's a good PR to start with. 

I think we shouldn't use a key formatter. As a comparison, we don't use a name formatter, because we rely on caption. Plus there is no concept in OLAP related to a formatted version of names and keys. Only captions and cells. So I would recommend eliminating the KeyFormatter. I'm sorry you wrote all of this to be told later to scrap it. Let's wait for @mkambol 's opinion before throwing it away.

There is also a change to RolapMemberBase which I'm a little iffy about. Do we really need to change the caption from Object to String? This is a major change in the API and it has (as we can see in this PR) quite a few ripples following it. I think we should keep it an Object, unless there is a very reason to change it. What was the motivation to make that change? Was this intended to cleanup type casting and make everything statically typed, or is there a deeper motivation?

mchen-len-son:
@lucboudreau thanks for the comment.

RolapMemberBase.setCaption(Object o) was previously added as part of the current issue and has already been merged in #673 . Now, considering the other changes, I've removed that method.

mchen-len-son:
@lucboudreau 

As for KeyFormatter - I agree, let's remove it. But, anyway, we should do something with basic formatting of level itself.
To achive the same effect as with KeyFormatter and to stay within the existing schema - we probably can add a default MemberFormatter. But in this case we should somehow expose the Key from RolapMember or just unformatted caption object to the Member interface.

mchen-len-son:
@mkambol 
just a friendly reminder to take a look at the PR

mkambol:
Hey @mchen-len-son 
There's a lot I like about this refactor.  Extracting out formatting logic in the way you've done will make it much easier to assure correctness and consistency between the different formatting types.

I'm a bit confused, though.  Like Luc says above, there's no need for a key formatter.  There's also not a need for a separate captionColumnFormatter vs. a memberFormatter.  The member formatter *is* the caption formatter.  Formatting is really only relevant to the elements which will actually be displayed (caption/cell/properties, not name or key).  It looks like your changes apply the formatting to the caption in both SqlMemberSource and SqlTupleReader before creating the member (using the captionColumnFormatter), but downstream the formatting will be done when the caption is retrieved (MemberBase.getCaption).  So unless I'm mistaken caption formatting is actually being applied twice with your change (once w/ the captionColumnFormatter, once with the memberFormatter).  I would think we could eliminate your new captionColumnFormatter altogether and just use the memberFormatter, applying formatting at the point the caption is retrieved.  Introducing a default member formatter as you suggest still makes sense.

mchen-len-son:
@lucboudreau @mkambol 

guys, I've updated the PR
- removed KeyFormatter
- removed CaptionColumnFormatter
- added caption raw value to the rolap member
- exposed key or caption column raw value to the Member interface. This value is now used by default MemberFormatter

I'm worrying about adding a new method to Member interface - it will break clients' stuff who already extended it. Probably I should add a new extending interface like it said in http://wiki.pentaho.com/display/PEOpen/API+Guidelines, shouldn't I?

mchen-len-son:
@lucboudreau @mkambol 
guys, can you please take a look?
thank you.

src/main/mondrian/rolap/RolapMemberBase.java
```diff
@@ -155,6 +155,18 @@ public RolapMember getParentMember() {
         return (RolapMember) super.getParentMember();
     }
 
+    @Override
+    public Object getCaptionValue() {
+        if (captionValue != null) {
+            return captionValue;
+        }
+        return getKey();
```
mkambol: I think the current behavior is that if captionColumn is not defined, we fall back first to name, then to finally to key.

mkambol:
@mchen-len-son I'm a little wary of modifying the interface, too.
I think storing the caption value could make sense.  As things stand right now, I don't think there's any way for a user defined member formatter to format the raw caption value.  In fact, I think if someone defined a formatter which attempted to .getCaption we'd end up w/ a stack overflow.
On the other hand-- maybe it would be cleaner to do the formatting at the point the member is created, avoiding interface changes.  AFAICT, the format shouldn't ever change after the fact, so there's no hard requirement that it has to be formatted on retrieval.  
@lucboudreau Do you have an opinion?

lucboudreau:
If we can avoid an API change, I'm all for it. And no, the format cannot change after it is defined in the schema. (That said, it could change at the query level, but that's post-processed)

mchen-len-son:
@lucboudreau @mkambol 

Guys, updated the PR.

Eventually I didn't touch the API, but couldn't get rid of "instanceof" using, unfortunately.

I also didn't move the member caption formatting to the point where member is created, because in this way, I would've had to either expose "this" in member constructor (member formatter requires the member object), or to write factory methods for the whole member hierarchy (even so, it wouldn't be cool, 'cause we don't have a full control on the hierarchy - members are not finalized and there are extending classes on clients side).

However, the changes are tested and the issue is fixed.

mchen-len-son:
@lucboudreau @mkambol 
Could you guys please take a look? Thanks!

mchen-len-son:
@lucboudreau @mkambol 
Guys, could you please take a look at the PR?
Thanks.

mkambol:
@mchen-len-son I'm seeing a number of test failures with your change.

E.g.

testCalcMemberCustomFormatterInQuery
testCalcMemberCustomFormatterInSchema
testCalcMemberCustomFormatterInQueryNegative

mchen-len-son:
Thanks, Matt.
Removed all the changes to classes constructors, as a breaking ones.

Now the tests are passing:
http://10.177.176.33:8081/view/MONDRIAN/job/mondrian_mchen-len-son/14/

mchen-len-son:
@lucboudreau @mkambol please take a look at the PR

mchen-len-son:
@lucboudreau @mkambol 
did you guys have a chance to take a look at the PR?

mchen-len-son:

PR #685, opened 2016-04-06
[SP-2643][MONDRIAN-2098] Alias error received when using an SQL Expre…

SergeyTravin:
…ssion in Mondrian Schema against a Greenplum database


SergeyTravin:
It is backport for https://github.com/pentaho/mondrian/pull/651

PR #688, opened 2016-04-11
[ANALYZER-3256] - NullPointerException when using Virtual Cube Drillt…

andrei-viaryshka:
…hrough with exclude filter on level not present in report

Changes:
* fix NPE;
* avoid adding empty where to sql query;
* unit test.


andrei-viaryshka:
@kurtwalker @lucboudreau @mkambol Could you please review?

brosander:
Base build failed with return code 1

lucboudreau:
I'm not too hot about this PR. It seems to avoid the presence of nulls, without really explaining them. Maybe if you explain why the columns came out null in those cases, we'd be more inclined to accept this PR.

PR #690, opened 2016-04-26
[CLEANUP] Exclude xercesImpl from olap4j

lgrill-pentaho:
Olap4j uses version 2.11.0, can evict 2.9.1 and cause build failure because current Mondrian code will not compile with XercesImpl 2.11.0


brosander:
Base build failed with return code 1

PR #691, opened 2016-04-28
[BACKLOG-7748] Add capital M for month number without leading zero

DFieldFL:


DFieldFL:
@rfellows @pminutillo @kurtwalker

PR #692, opened 2016-05-11
[MONDRIAN-2462] arrayIndexOutOfBounds encountered when performing rollup

andrey-markelov:
@mkambol please review

### DIFF
--- a/src/main/mondrian/rolap/agg/SegmentBuilder.java
+++ b/src/main/mondrian/rolap/agg/SegmentBuilder.java
@@ -4,7 +4,7 @@
 // http://www.eclipse.org/legal/epl-v10.html.
 // You must accept the terms of that agreement to use this software.
 //
-// Copyright (c) 2002-2015 Pentaho Corporation..  All rights reserved.
+// Copyright (c) 2002-2016 Pentaho Corporation..  All rights reserved.
 */
 package mondrian.rolap.agg;
 
@@ -23,7 +23,6 @@ import org.apache.log4j.Logger;
 import org.olap4j.impl.UnmodifiableArrayList;
 
 import java.math.BigInteger;
-
 import java.util.*;
 import java.util.Map.Entry;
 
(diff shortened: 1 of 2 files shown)


lucboudreau:
I'm not sure I see the bug here. Which line fixes it exactly? I don't see it. 

I see an array being converted into a list, and it will probably return null instead of throwing an exception, but it does not seem to fix an issue as far as I can tell.

andrey-markelov:
@lucboudreau the bug was in incorrect processing of input constraint columns, that contains duplicated expressions. It related to impossibility of previous definition of "axes" size.

lucboudreau:
I'm not sure the test makes sense either. You're passing two columns in ``keepColumns``, but you're assert is for 3 columns returned. @mkambol  Am I missing something? You're more familiar with this code than I am.

mkambol:
Looking back at the test in the original case that prompted this change, I see that there are some unusual features of the customer's schema.  The schema has at least one dimension where two levels have the same key value, but one of the two has a different column specified as the nameColumn.  AFAICT there's nothing in principal wrong with this (although it's a little odd), but it results in a constrainedColumn collection for the rollup which has a duplicate column name present.  The SegmentBuilder.rollup logic has an implicit assumption that the columns are unique.

Here's a functional test that reproduces the issue:  https://gist.github.com/mkambol/750c937d467a35bfc62026216d7c26b1

@andrey-markelov's change gets around the problem by allowing for duplicate column names to be present in the constraint.  That seems like a reasonable option.  I'd want to make sure that this duplicate column issue isn't masking other problems, though-- it seems possible that having duplicate columns present in the constraint of a segment could potentially cause problems elsewhere.  @lucboudreau, does anything occur to you off-hand?  

BTW- I'd expect that simply removing the nameColumn specified in the schema (or making sure it's not doing double duty as a key) would workaround the error.

lucboudreau:
Segments were meant to be as lean as possible. This means that we would add a whole duplicate column for no reason. I think the code which uses Segments should read the same column twice when possible.

lucboudreau:
Your PR has 3 errors and 8 failures. Please fix those unit tests, or provide a fix which does not cause these regressions.

andrey-markelov:
@lucboudreau could you please provide me more detailed error info? It have successfully built on my local machine using `ant clean-deep jar`

lucboudreau:
My bad. I've ran the tests under Java 8. Seems everything is OK with it.

PR #693, opened 2016-05-14
[MONDRIAN-2434] Drill-through filtering not properly working when dim…

stanislau-strelchanka:
[MONDRIAN-2434] Drill-through filtering not properly working when dimension level is excluded from filter and not displayed on canvas


stanislau-strelchanka:

PR #694, opened 2016-05-16
[MONDRIAN-2471] fix for processing sql columns containing dots

andrey-markelov:
@mkambol @kurtwalker please merge

### DIFF
--- a/src/main/mondrian/spi/impl/JdbcDialectImpl.java
+++ b/src/main/mondrian/spi/impl/JdbcDialectImpl.java
@@ -4,7 +4,7 @@
 // http://www.eclipse.org/legal/epl-v10.html.
 // You must accept the terms of that agreement to use this software.
 //
-// Copyright (c) 2002-2014 Pentaho Corporation..  All rights reserved.
+// Copyright (c) 2002-2016 Pentaho Corporation..  All rights reserved.
 */
 package mondrian.spi.impl;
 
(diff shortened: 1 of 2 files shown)


mkambol:
Hey @andrey-markelov, 
I think this change is safe, but last we chatted about this one there was an open question about whether there is any need to support table references like "ower.table".  I believe the answer to that is no, since the "schema" attribute on Table should do the same thing more cleanly, but it would be nice to know there's not a use case we aren't thinking of.  Have you learned any more since last we exchanged emails?

andrey-markelov:
@mkambol
Hi Matt. I am not absolutely sure that this old feature is being used somewhere, but I could not find for any usages of this form

src/main/mondrian/spi/impl/JdbcDialectImpl.java
```diff
@@ -331,26 +331,10 @@ public void quoteIdentifier(final String val, final StringBuilder buf) {
             return;
         }
 
-        int k = val.indexOf('.');
```
mbatchelor: This was purposely put in to handle "owner.table". and replacing that with "owner"."table". based on this commit:
https://github.com/pentaho/mondrian/commit/2930d1256605148dbc7e2641e7f0b8904b5d7efc#diff-26f86eb9f6e0b15e1c4af7f0457b0801

Do your test cases cover that ?

mkambol: @mbatchelor This logic scanning for the '.' is pretty ancient, and I think it pre-dates the "schema" attribute, which would be the preferred way to specify the "owner".  Given that periods are allowed characters in table and column names in many databases, we can't safely extract the schema part based on the presence of the dot.

Including @lucboudreau in case he has any concerns.

PR #698, opened 2016-05-26
[MONDRIAN-2126] Wrong unique member name obtained through XMLA when a…

stanislau-strelchanka:
[MONDRIAN-2126] Wrong unique member name obtained through XMLA when alias are used w/ shared dimensions - override method

### DIFF
--- a/src/main/mondrian/rolap/RolapCubeMember.java
+++ b/src/main/mondrian/rolap/RolapCubeMember.java
@@ -5,10 +5,9 @@
 // You must accept the terms of that agreement to use this software.
 //
 // Copyright (C) 2001-2005 Julian Hyde
-// Copyright (C) 2005-2011 Pentaho and others
+// Copyright (C) 2005-2016 Pentaho and others
 // All Rights Reserved.
 */
-
 package mondrian.rolap;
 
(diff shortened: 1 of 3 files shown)


stanislau-strelchanka:

src/main/mondrian/rolap/RolapCubeMember.java
```diff
@@ -201,6 +200,10 @@ public Object getPropertyValue(String propertyName, boolean matchCase) {
         return member.getPropertyValue(propertyName, matchCase);
     }
 
+    public Object getPropertyValue(String propertyName) {
```
kurtwalker: Where is this new method called from?

stanislau-strelchanka: It is called from [here].(https://github.com/pentaho/mondrian/blob/master/src/main/mondrian/xmla/XmlaHandler.java#L2277)
There is an overridden [method](https://github.com/pentaho/mondrian/blob/master/src/main/mondrian/rolap/RolapCubeMember.java#L166) in RolapCubeMember, but it's overloaded version existed only in base class - DelegatingRolapMember.

PR #711, opened 2016-06-08
MONDRIAN-2497 - Fix mondrian test failures

stanislau-strelchanka:
MONDRIAN-2497 - fix xmla basic test


testsrc/main/mondrian/xmla/XmlaBasicTest.java
```diff
@@ -744,12 +744,12 @@ public void testExecuteAliasWithSharedDimension()
             + "<Schema name=\"foodmart-xmla-alias-bug\">\n"
             + "  <Dimension name=\"Customers\">\n"
             + "    <Hierarchy hasAll=\"true\" allMemberName=\"All Customers\" primaryKey=\"customer_id\">\n"
-            + "      <Table schema=\"foodmart\" name=\"customer\"/>\n"
+            + "      <Table name=\"customer\"/>\n"
```
mkambol: @mchen-len-son Why does removing the schema attribute fix the tests?  Can you provide a little background on what was failing and why?

mchen-len-son: @mkambol 
it was failing for Oracle, 'cause in Oracle users and schemas are the same thing, and schema attribute is not filled then. 
Tested for Oracle and Mysql.

stanislau-strelchanka: @mkambol 
When schema attribute is used, mondrian will generate sql which runs query against 'foodmart.customer' table which is ok for mysql, but not for Oracle. Otherwise, if this attribute is not used, it will infer schema from connection properties and create correct query for both databases. I've checked foodmart and steelwheels shemas: they don't contain this attribute in table element. I guess it was a copy paste error from some bad example.

PR #712, opened 2016-06-08
[MONDRIAN-2496] - Parent/child use captionColumn

martijn-basesoft:
Fix with testcase for MONDRIAN-2496

### DIFF
--- a/src/main/mondrian/rolap/SqlMemberSource.java
+++ b/src/main/mondrian/rolap/SqlMemberSource.java
@@ -1385,6 +1385,10 @@ RME is this right
         public int getOrdinal() {
             return dataMember.getOrdinal();
         }
+
+        public String getCaption() {
+            return dataMember.getCaption();
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


kurtwalker:
if you could change your commit message to the format "[MONDRIAN-2496] - Parent/child use captionColumn" that will help our tracking.  thanks.

martijn-basesoft:
done!

martijn-basesoft:
I noticed the new RolapMemberBase was created after a .setCaption was called, but that information isn't passed. I moved the .setCaption() after the new member is created so all members get that call now.  This makes the failing tests pass for me again. What do you think of this change?

kurtwalker:
I don't follow how this would change anything since it was previously being set on the dataMember which is ultimately what's used to get the caption anyway.  Can you help me understand?  I'm hesitant to merge because it has a potential to affect other code paths.

martijn-basesoft:
It seems if we override getCaption() in RolapParentChildMember then the xmla tests executeHR seem to fail because they use a closure table on the HR Cube that has a nameColumn (and no captionColumn) set. 

With getCaption() overriden, it then calls dataMember.getCaption() which (because caption equals null) calls getName() in OlapElementBase. It then returns the value (integer) in these xmla tests, and thats wrong. 

I noticed if I move the .setCaption() after the parent-child member is created then the getName() behaviour is equal in both cases and no getCaption() override is needed. 

We can keep the .getCaption() override but I'll need a hint how to fix the getName() issue in case of nameColumn set. Any hints where to look for?

PR #714, opened 2016-06-14
[ANALYZER-3256] - NullPointerException when using Virtual Cube Drillthrough with exclude filter on level not present in report

andrei-viaryshka:
Add method to identify appropriate base cube in order to get correct star key column.


andrei-viaryshka:
@kurtwalker @lucboudreau @mkambol Could you please review?

PR #715, opened 2016-06-17
[MONDRIAN-1599]  Virt Cube MDX mishandled w/ non-joining dims

mkambol:
Adding support for native evaluation of sets in the context of a
virtual cube in which no cubes "fully join".  That is, if there is no
cube or cubes which are applicable to all dimensions in the native
targets.  For example, with a crossjoin of Warehouse and Gender, no
single cube joins to both.  Formerly native eval logic depended on the
presence of one or more fully joining cubes in order to find tuples,
with the assumption that the absence of a fully joining cube implies
that the resulting set will be empty.  That assumption is incorrect
when ValidMeasure is use (or other MDX expressions which changes
dimensional context).

This commit takes the approach of "grouping" native targets based on
their applicable cubes.  So with the above example, Gender and
Warehouse are grouped and evaluated separately, then crossjoined after
the fact.  This is more expensive than the former logic, which
executed a single UNION'd SQL query with virtual cubes, so to avoid
unnecessary execution this first checks whether one or more
measures could shift context.  Otherwise, if there is no fully joining
cube, then it's safe to return an empty set.

http://jira.pentaho.com/browse/MONDRIAN-1599
http://jira.pentaho.com/browse/MONDRIAN-2280

### DIFF
--- a/src/main/mondrian/olap/fun/CrossJoinFunDef.java
+++ b/src/main/mondrian/olap/fun/CrossJoinFunDef.java
@@ -17,6 +17,7 @@ import mondrian.olap.*;
 import mondrian.olap.type.*;
 import mondrian.resource.MondrianResource;
 import mondrian.rolap.RolapEvaluator;
+import mondrian.rolap.SqlConstraintUtils;
 import mondrian.server.Execution;
 import mondrian.server.Locus;
 import mondrian.util.CancellationChecker;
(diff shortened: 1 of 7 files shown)


mkambol:
@lucboudreau 

I updated based on our review to use the cubes associated with the query first when determining target grouping, rather than all base cubes in the virtual cube.  Most other changes have been general cleanup and doc.

I've run the test suite native / non-native, agg / no-agg.

PR #716, opened 2016-06-21
[MONDRIAN-1803] - Getting Wrong Results on 3.5 while correct results …

stanislau-strelchanka:
[MONDRIAN-1803] - Getting Wrong Results on 3.5 while correct resultson 3.4 for same MDX and schema properties also

### DIFF
--- a/src/main/mondrian/resource/MondrianResource.xml
+++ b/src/main/mondrian/resource/MondrianResource.xml
@@ -883,6 +883,12 @@
     </text>
 </message>
 
+<message id="8000200" name="MembersQuantityExceedsMaxConstraints">
+    <text>
+        Quantify of supported calculated members(''{0}'') exceeds value of the mondrian.rolap.maxConstraints property.
+    </text>
+</message>
+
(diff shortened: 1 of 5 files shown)


stanislau-strelchanka:
@kurtwalker, @lucboudreau, @mkambol, @mchen-len-son, please review

stanislau-strelchanka:
@kurtwalker, @lucboudreau, @mkambol, @mchen-len-son, @pamval  please review

lucboudreau:
@stanislau-strelchanka Could you explain what this fixes? From reading the code, I'm guessing (maybe I'm wrong) that this case is about a large number of calculated members tripping an optimization procedure and removing a set of predicates. Is this exact?

stanislau-strelchanka:
@lucboudreau yes, it avoids using native filter in case there will be too many args in 'in' clause. Otherwise the will be exception and incorrect result will be returned.

stanislau-strelchanka:
found broken tests, will close pr until find the reason

PR #717, opened 2016-06-22
[MONDRIAN-2476] deadlock encountered between DynamicContentFinder and…

mchen-len-son:
… distributed cache

- moved schema clean up out of critical section

Still could not reproduce the issue. Putting here the changes proposed by customer in JIRA.
I checked them - they look fine to me, it's indeed should help to avoid the situation they described in the case.


mchen-len-son:

mchen-len-son:
@lucboudreau @mkambol 
Guys, could you please take a look at the PR?
Thanks.

PR #718, opened 2016-06-22
[MONDRIAN-2358] fixed incompatibility of calculated and cached result

andrey-markelov:
@mkambol @lucboudreau could you review it?

### DIFF
--- a/src/main/mondrian/rolap/RolapCubeHierarchy.java
+++ b/src/main/mondrian/rolap/RolapCubeHierarchy.java
@@ -5,13 +5,18 @@
 // You must accept the terms of that agreement to use this software.
 //
 // Copyright (C) 2001-2005 Julian Hyde
-// Copyright (C) 2005-2013 Pentaho and others
+// Copyright (C) 2005-2016 Pentaho and others
 // All Rights Reserved.
 */
-
 package mondrian.rolap;
 
-import mondrian.olap.*;
+import mondrian.olap.Access;
+import mondrian.olap.Formula;
+import mondrian.olap.Level;
+import mondrian.olap.Member;
+import mondrian.olap.MondrianDef;
+import mondrian.olap.MondrianProperties;
+import mondrian.olap.Util;
 import mondrian.olap.fun.VisualTotalsFunDef;
 import mondrian.rolap.TupleReader.MemberBuilder;
 import mondrian.rolap.sql.MemberChildrenConstraint;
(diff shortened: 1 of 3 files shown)


mkambol:
@andrey-markelov The problem described with this bug is that failing to have a 1:1 relationship between ordinalColumn and level name/key is an issue with the schema and data.  If you have a member that has an ordinal of both 1 & 3 in the dim table, how do you sort it with respect to a member with ordinal 2?  Right now the behavior is indeterminate, but the bottom line is Mondrian can't know the order, and shouldn't try to guess.

The suggestion I made in the comment of the jira case was to try to check whether the same member is found to have 2 distinct ordinalColumn values when loading in SqlTupleReader and SqlMemberSource.  If this happens we could then minimally log an ERROR, or throw an exception.

andrey-markelov:
@mkambol you're right, I'll add logging for this case, I missed this point

But also there is a problem that related to this situation: while expanding child members of duplicates records list - a different result when we calculating child members and when loading. Is this abnormally to process the duplicated members?

andrey-markelov:
@mkambol 
Hi Matt,
I've add logging to code block that you pointed in comment
https://github.com/pentaho/mondrian/pull/718/commits/434489630f1ef767a946de8e6fa76137f9bada86

but not sure that it's right.
How could we check that we processing non 1:1 relationship ?

mkambol:
Hey @andrey-markelov.  The comment I made in the case was this: 

> One option might be to keep a temporary hash of members to ordinal values in SqlTupleReader.internalAddRow. Rather then skipping over the ordinal column value when loading members, we could compare the value to what’s been stored here:
> https://github.com/pentaho/mondrian/blob/master/src/main/mondrian/rolap/SqlTupleReader.java#L236
> We could then log if we’ve been told 2 different sort values for the same member (or maybe better, throw an exception).

Simply warning every time ordinal and key are different is not correct, since in most cases that's probably just fine.  The problem is when the same member is associated with more than one ordinal value.

Also note that my other comment in that same case mentioned that we should look at SqlMemberSource, as well, since the same issue could be hit there.

andrey-markelov:
@mkambol thanks, but I saw this comment

I see only "month_of_year", where is the second ordinal column?
As I understood, the problem was in one-to-many relation b/w level expression(quarter) and ordinal exp(month), not it so?

mkambol:
@andrey-markelov 
Consider a case where the [Month] level is defined to have column="month", but ordinalColumn="day_of_month".  In this case [January] will have ordinal values {1, 2, 3, ...}, rather than a single fixed value.  [February] will have {1,2,3,...} as well, and determining ordinal position for the two members is impossible since the values overlap.

That's an extreme case, but we've seen many more subtle cases where an ordinalColumn has multiple values for the same member value.  We need to check for this while loading members in SqlTupleReader and SqlMemberSource and give an error if we detect the problem.

PR #719, opened 2016-06-29
[MONDRIAN-2476] deadlock encountered between DynamicContentFinder and…

mchen-len-son:
… distributed cache

- fixed possible NPE


mchen-len-son:
@lucboudreau @mkambol please take a look

PR #721, opened 2016-06-30
[SP-2744][MONDRIAN-2476] deadlock encountered between DynamicContentF…

mchen-len-son:
…inder and distributed cache

- moved schema clean up out of critical section


mchen-len-son:
@lucboudreau @pamval please review

mchen-len-son:
@lucboudreau @pamval can you please take a look?

PR #726, opened 2016-07-01
[SP-2738][ANALYZER-3256] - NullPointerException when using Virtual Cu…

mchen-len-son:
…be Drillthrough with exclude filter on level not present in report

This is a backport of #714


mchen-len-son:
@kurtwalker @mkambol please review

PR #728, opened 2016-07-01
[MONDRIAN-2502] - Emit dialect name into logs

mbatchelor:


lucboudreau:
Marc,

This will not have the effect you think it does. We create all of the dialects, then use a Visitor-like approach ( see accept(java.sql.Connection) ).

Also, please remove the @Override annotations.

mbatchelor:
According to Luc, we don't just instance the dialect we need, we instance them all. I'll need a different approach.

PR #733, opened 2016-07-15
[TEST] Fixes SegmentHeader portability issues across JVMs. The column…

lucboudreau:
…s and regions are now sorted when building the hash code and ID, thus allowing different JVMs sharing the same cache. We can't sort the columns directly, as they are populated according to SQL specs, and must always stay in sync with the SegmentDataset it represents, but we sort when displaying a string representation, or calculating hashes, which happens only once per header's lifespan.


lucboudreau:
@mkambol @kurtwalker @tgf When you guys get a chance to review this, it should take care of all problems w/ SegmentHeader and JVM versions.

PR #738, opened 2016-07-24
[MONDRIAN-2494] Single timer clears serverInfo of each FileRepository…

pentaho-nbaker:
… in play. Shutdown, Destroy and a WeakHashMap all ensure the instance Set contains valid FileRepository instances only.


pentaho-nbaker:
@lucboudreau no idea why it would fail to find the "resolve" task

lucboudreau:
Wingman and mondrian don't like each other (yet) so don't expect it to work here.

PR #756, opened 2016-09-07
[BISERVER-7257] - Support for MariaDB as a DS

mbatchelor:


mbatchelor:
7.1 was reorganized for Maven. This had to be completely redone - resubmitted as #786

PR #758, opened 2016-09-07
[BACKLOG-10323] Standardizing on javassist version 3.20.0-GA

pdesai16:
- The Javassist version 3.20.0-GA is stored as bundle in maven
      	   repo whereas javassist version 3.12.1.GA was stored as jar.
	   Changing the build.xml to use type as either jar or bundle while
	   retrieving from repo so that it can retrieve the new version of javassist.


pedrofvteixeira:
👍

PR #760, opened 2016-09-16
[MONDRIAN-1803] - check for supported query

stanislau-strelchanka:


stanislau-strelchanka:

mkambol:
@stanislau-strelchanka -  I like where this is going.  I'm not sure about adding the .isSupported method to SqlQuery, though.  The query itself actually *is* supported-- we exclude the big IN constraint in the query.  It's the constraint that's not supported.

Since it's the adding of the constraint that is unsupportable, I wonder if we could keep this on the constraint class.  In this case we're hitting an unsupportable condition when applying the singleValueInList in SqlConstraintUtils.  Maybe the constraint itself could be passed down that chain, and marked as unsupportable if this (or anything else) is encountered which would result in invalid SQL.

Also, you've added to the FilterConstraint class, but I wonder if the .isSupported should be part of a higher level interface.  Actually, TopCountConstraint has a .isValid method that serves a similar purpose.  Maybe we should pull that up into the SetConstraint class and have FilterConstraint implement it.

PR #771, opened 2016-10-24
[PPP-3506] - XXE Security Vulnerability in xml parsers

YuryBY:
@mkambol, @mbatchelor, could please take a look?  

Also, I'm not sure if we should do the same chabges here: https://github.com/pentaho/mondrian/blob/53295633f56fd13290169e3271bbab489ef4e3a6/src/main/mondrian/xmla/impl/DefaultXmlaServlet.java#L68 (is XXE possible there)? Thanks.

### DIFF
--- a/src/main/mondrian/tui/XmlUtil.java
+++ b/src/main/mondrian/tui/XmlUtil.java
@@ -9,6 +9,7 @@
 package mondrian.tui;
 
 import mondrian.olap.Util;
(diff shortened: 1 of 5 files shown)


mkambol:
@YuryBY - yes, I think you should also address the usage of DocumentBuilderFactory in DefaultXmlaServlet.

PR #780, opened 2016-11-09
[BACKLOG-8047] - splitting workbench assembly into a separate sub-module

kurtwalker:


kurtwalker:
don't merge, wait for @smaring or @lgrill-pentaho

lucboudreau:
Sorry Kurt, you just hit the line endings bug in Wingman. You can always merge it and then trigger some other which does not touch these files. Or alternatively, you can try your hand at finding why Wingman's Git client decides to change the line endings. We tried all sorts of flags like "-Xtheirs, -Xignore-space-change" but nothing seems to affect it.

PR #782, opened 2016-11-10
[PPP-3506] - XXE Security Vulnerability in xml parsers

YuryBY:
- eigenbase-xom dependency is updated to 1.3.5

@lucboudreau, let's waiti for wingman since have some issues while building and not sure if it relates to local environment. Thanks.


lucboudreau:
@YuryBY Wingman won't build this. It can't. What issues did you run into?

PR #786, opened 2016-11-14
[BISERVER-7257] - Support for MariaDB as a DS

mbatchelor:


mbatchelor:
Wingman failures aren't caused by this commit. Merging pull request and will watch devci. Note that this compiles locally.

PR #799, opened 2016-12-19
Fix for jira issue MONDRIAN-2542

truvenganong:
Prevent logging of JdbcPassword when there is a
MemoryLimitExceededException.


mbatchelor:
@truvenganong - Thank-you so much for the contribution!

PR #800, opened 2016-12-19
fix snapshot build on devci

stanislau-strelchanka:


stanislau-strelchanka:
@lucboudreau @mkambol @lgrill-pentaho please review

PR #804, opened 2017-01-06
[MONDRIAN-2442] Mondrian throws NullPointerException with report that…

stanislau-strelchanka:
[MONDRIAN-2442] Mondrian throws NullPointerException with report that uses aggregate tables - ignore name column exp if nameColumn and column are the same

### DIFF
--- a/mondrian/src/it/java/mondrian/test/BasicQueryTest.java
+++ b/mondrian/src/it/java/mondrian/test/BasicQueryTest.java
@@ -5,7 +5,7 @@
 // You must accept the terms of that agreement to use this software.
 //
 // Copyright (C) 2003-2005 Julian Hyde
-// Copyright (C) 2005-2016 Pentaho
+// Copyright (C) 2005-2017 Pentaho
 // All Rights Reserved.
 //
 // jhyde, Feb 14, 2003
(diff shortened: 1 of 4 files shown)


stanislau-strelchanka:
The problem is happening when there are same 'column' and 'columnName' attributes in Level element.
When we are looking for a column for aggUsage.rColumn, named column is picked up instead of key column.
Not creating a named column solves the issue.

stanislau-strelchanka:
The problem is happening when there are same 'column' and 'columnName' attributes in Level element.
When we are looking for a column for aggUsage.rColumn, named column is picked up instead of key column.
Not creating a named column solves the issue.

@lucboudreau @mkambol @kurtwalker please review

kurtwalker:
why are there so many existing test failures?  Didn't we have it down to zero not long ago?  I'd prefer to fix all those up first if possible.

stanislau-strelchanka:
I've sent a test pull request: #805 without any changes. And there were failing tests.
Checked them locally, only one was failing because of my commit - https://github.com/pentaho/mondrian/blob/c55bc99a7be5f0f24a5af6d73c26e0c0f6e8da96/mondrian/src/it/java/mondrian/test/DrillThroughTest.java#L852-L993.
I've updated expected result, as there will be no extra column if 'column' and 'nameColumn'
attributes are the same.

@lucboudreau @mkambol @kurtwalker please review

kurtwalker:
Do you know when these tests started failing?  I know it's not this commit, but it's still a concern.

For the one newly broken test you wrote, "I've updated expected result".  Did you forget to push another commit to this PR?

stanislau-strelchanka:
Regarding tests with not matching segment ID's: I recall that it is old problem and encountered it few
times after mavenization. 

I've merged two commits. 'testBug1438285' was failing and I removed two lines from 'expectedSql'.

@lucboudreau @mkambol @kurtwalker please review

stanislau-strelchanka:
@lucboudreau @mkambol @kurtwalker please review

PR #807, opened 2017-01-10
[MONDRIAN-2358] If ordinalColumn does not have a 1:1 relation w/ leve…

stanislau-strelchanka:
[MONDRIAN-2358] If ordinalColumn does not have a 1:1 relation w/ level name, no error is given and inconsistent results can occur


stanislau-strelchanka:
@mkambol @kurtwalker @lucboudreau @mchen-len-son plrease review

PR #810, opened 2017-01-13
[MONDRIAN-2528] Restrictions applied to measures in a schema does not…

ivanpogodin:
… work as intended in BA 6.1.0.1


ivanpogodin:
@mbatchelor , @kurtwalker , could you review it, please?
Two latest Wingman launches are on the same code.

ivanpogodin:
@mbatchelor , @kurtwalker , @pamval , could you review it, please?
Squashed

kurtwalker:
I'd like to see an integration test in SchemaTest that creates a schema that demonstrates the case we are fixing here.  There are several role tests inside SchemaTest that you can use for an example.

PR #826, opened 2017-02-06
[MONDRIAN-2285] Native filter with aggregate tables causes NPE when N…

stanislau-strelchanka:
[MONDRIAN-2285] Native filter with aggregate tables causes NPE when Name column is not the same as key column - check another fix


stanislau-strelchanka:
Exception is thrown when constraining members having 'nameColumn' attr set to different column than 'column' attr.

@mkambol @lucboudreau @kurtwalker @mchen-len-son can you please review it

PR #831, opened 2017-02-13
[MONDRIAN-2245] Mondrian omits primaryKeyTable if it is the right table of a Join

Advent51:
If primaryKeyTable do not references to the leftmost table, then we swap it.


Advent51:
All these tests passed locally.

lucboudreau:
I tested the PR locally and I also get all tests passing. @mkambol Could you do a quick local run as well? Just tryin to figure out why that happens on wingman..

Advent51:
@lucboudreau @mkambol So, what about the fix itself? Could you merge it?

PR #833, opened 2017-02-14
[TEST] Failing tests

Advent51:
Test was updated and now passes succesfully


Advent51:
That is strange that this fix brakes this test: mondrian.rolap.NativeEvalVirtualCubeTest.testShouldUseCache
Because it was broken on my local machines with Linux and Windows and was fixed by these changes.

Advent51:
The problem is that we use embedded-MySQL which is version 5.6.14, but we should use 5.7.5 or later one.

PR #837, opened 2017-02-15
[ANALYZER-3283] - Levels not accessible are displayed in fields list

YuryBY:


lucboudreau:
Let's not make the getAccess method part of the public API. Can we keep it as a package protected (or even class private) method?

YuryBY:
@lucboudreau, @mkambol, I've just provided test that demonstrates the issue. Regarding decreasing of getAccess visibility. Currently it's impossible since https://github.com/pentaho/mondrian/blob/1712450ac2b25b778fedb729639816bb8941c652/mondrian/src/main/java/mondrian/olap/Role.java#L55 is an interface. But I'll do it if you suggest another place where this logic can be implemented. Thanks.

lucboudreau:
I can't quite come up with an alternative. The problem I foresee with the solution is that it changes a public API, which a lot of customers (OEMs) extend. It means we can only check this solution in the next release, and we cannot backport it.

Please check if you can come up with any alternative. Let us know if not and we'll get together to figure this out.

YuryBY:
@lucboudreau, @mkambol, PR was updated. Current approach should not break anything on the side and can be backported. Thanks.

lucboudreau:
@YuryBY Your solution should work. If an OEM injects his own role, the UnionRole class will not be used and the getAccess method will never be called. I still don't like it being public, in case someone mistakenly calls it, but it will do for now. It's more of a JVM limitation TBH.

@mkambol - please do a second review. I think it's safe, but would like your opinion.

YuryBY:
@mkambol, any updates?

PR #839, opened 2017-02-17
[TEST] Failing tests

Advent51:
JDK dependent tests were updated - now JDK version doesn't cause any headache;
Some tests were result order dependent despite it wasn't necessary - it caused another OS/JDK dependency.


Advent51:

PR #827, opened 2017-02-08
[MONDRIAN-2530] Native Filter can exclude null members

Advent51:
fixed regression BACKLOG-14579

PR #796, opened 2016-12-06
[MONDRIAN-2530] Native Filter can exclude null members

Advent51:
-fixed
-this is a second PR with similar changes, I did it because I need a wingman's check, but it failed to build on previous PR.
-please don't merge before wingman's checks pass.

PR #841, opened 2017-02-17
[SP-3283] Backport of MONDRIAN-2530 - Native Filter can exclude null members (7.0 Suite)

Advent51:
Backport of #796 with #827


Advent51:

PR #843, opened 2017-02-17
[SP-3284] Backport of MONDRIAN-2530 - Native Filter can exclude null members (6.1 Suite)

Advent51:
Backport of #796 with #827


Advent51:

PR #844, opened 2017-02-18
[MONDRIAN-2399] No way to use an AvgFromSum calculation with explicit…

stanislau-strelchanka:
[MONDRIAN-2399] No way to use an AvgFromSum calculation with explicitly defined aggregate tables


stanislau-strelchanka:
@mchen-len-son @mkambol @kurtwalker @lucboudreau please review

PR #846, opened 2017-02-28
[TEST] - changing embedded mysql database to use memory storage engine

kurtwalker:
Locally full IT suite went from 14 mins to 8:42.  woohoo.


kurtwalker:
@lucboudreau

PR #847, opened 2017-03-06
[MONDRIAN-625] The Descendants function will only return Non Empty member children, even if it is not specified inside of NON EMPTY

Advent51:
Descendants function is NON EMPTY dependent now.

### DIFF
--- a/mondrian/src/it/java/mondrian/rolap/NonEmptyTest.java
+++ b/mondrian/src/it/java/mondrian/rolap/NonEmptyTest.java
@@ -4296,7 +4296,7 @@ public class NonEmptyTest extends BatchTestCase {
         assertQueryReturns(
             "WITH SET [#DataSet#] AS 'Crossjoin({Descendants([Customers].[All Customers], 2)}, {[Product].[All Products]})' \n"
             + "SELECT {[Measures].[Unit Sales], [Measures].[Store Sales]} on columns, \n"
-            + "Hierarchize({[#DataSet#]}) on rows FROM [Sales]",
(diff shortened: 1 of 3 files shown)


Advent51:
Well, the problem with `NonEmptyTest.java` modification still persists: last wingman fail was caused by the modification of `testBugMondrian321` in `NonEmptyTest.java`

When I fixed the behavior of `descendantsByDepth` function, this test failed. But if we look at [MONDRIAN-321](http://jira.pentaho.com/browse/MONDRIAN-321) we can see that the previous expected result was wrong, because `NON EMPTY` wasn't specified on axis. (Without `NON EMPTY` but with my fix it returns result set with empty members as expected)

Advent51:

mkambol:
@Advent51 I think you're right-- testBugMondrian321 appears to be testing the opposite of what it's supposed to.
Can you fix that test?  It really should do something like verifySameNativeAndNot().

Advent51:
@mkambol Actually I did that. Look at [this](https://github.com/pentaho/mondrian/pull/847/files#diff-b2ced3a3d22109e06eb8b00f4075bf60R4299)

mkambol:
Sorry @Advent51, I don't follow.  Does testBugMondrian321 succeed or fail?  

Either way, that test seems like a good candidate to verify same results native and not.  .verifySameNativeAndNot() gives you a good way to do that.

Advent51:
@mkambol please review.
I added `verifySameNativeAndNot()` test as you requested.

Advent51:
@mkambol @lucboudreau please review

PR #848, opened 2017-03-06
[TEST] - removing main method from Main test class.  We should always…

kurtwalker:
… be using the TestSuite through a junit TestRunner


kurtwalker:
@emartin-pentaho @lucboudreau

kurtwalker:
failures are same ones we've been fighting with recently

PR #849, opened 2017-03-08
[CLEANUP] - resources should write to generated instead of target

kurtwalker:
This helps if you want intellij to compile to it's own out directory instead of target.


kurtwalker:
@pentaho/ackbar @lucboudreau

PR #851, opened 2017-03-13
[SP-3344] Backport of MONDRIAN-2399 - No way to use an AvgFromSum cal…

stanislau-strelchanka:
[SP-3344] Backport of MONDRIAN-2399 - No way to use an AvgFromSum calculation with explicitly defined aggregate tables (6.1 Suite)


stanislau-strelchanka:
@mkambol @kurtwalker @lucboudreau @mchen-len-son please review

PR #852, opened 2017-03-13
[SP-3345] Backport of MONDRIAN-2399 - No way to use an AvgFromSum cal…

stanislau-strelchanka:
[SP-3345] Backport of MONDRIAN-2399 - No way to use an AvgFromSum calculation with explicitly defined aggregate tables (7.0 Suite)


stanislau-strelchanka:
@mkambol @lucboudreau @kurtwalker @mchen-len-son please review

PR #859, opened 2017-03-23
[TEST]Failing test

Advent51:
Fix for the failing test


Advent51:
@mkambol @lucboudreau @kolinus please review

Advent51:
As we already discussed, we cannot fix this test in master, because we cannot update the MySQL server version in embedded-MySQL.
So I was asked to fix it at least in 3.13

PR #861, opened 2017-03-28
[MONDRIAN-2138] Incorrect results when non-all defaultMember is used with .children in NON EMPTY context

Advent51:
- added member constraint

### DIFF
--- a/mondrian/src/main/java/mondrian/rolap/SqlContextConstraint.java
+++ b/mondrian/src/main/java/mondrian/rolap/SqlContextConstraint.java
@@ -5,7 +5,7 @@
 // You must accept the terms of that agreement to use this software.
 //
 // Copyright (C) 2004-2005 TONBELLER AG
-// Copyright (C) 2006-2016 Pentaho and others
+// Copyright (C) 2006-2017 Pentaho and others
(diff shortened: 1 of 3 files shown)


mondrian/src/main/java/mondrian/rolap/SqlContextConstraint.java
```diff
@@ -282,10 +282,8 @@ public void addMemberConstraint(
             evaluator.restore(savepoint);
         }
 
-        // comment out addMemberConstraint here since constraint
-        // is already added by addContextConstraint
-        // SqlConstraintUtils.addMemberConstraint(
-        //        sqlQuery, baseCube, aggStar, parent, true);
+         SqlConstraintUtils.addMemberConstraint(
```
mkambol: Hey @Advent51 - just so I understand, why is the comment from before your change not valid?  I.e. is already added by addContextConstraint is already added by addContextConstraint?

Advent51: @mkambol well, at least I didn't found the addition of the "WHERE parent = value" constraint  in SqlConstraintUtils.addContextConstraint

PR #865, opened 2017-03-31
[MONDRIAN-2562] Mondrian can't parse query correctly in pentaho7

Advent51:
- fixed scheme name enclosing in quotation marks

### DIFF
--- a/mondrian/src/main/java/mondrian/spi/Dialect.java
+++ b/mondrian/src/main/java/mondrian/spi/Dialect.java
@@ -4,7 +4,7 @@
 * http://www.eclipse.org/legal/epl-v10.html.
 * You must accept the terms of that agreement to use this software.
 *
-* Copyright (c) 2002-2013 Pentaho Corporation..  All rights reserved.
+* Copyright (c) 2002-2017 Pentaho Corporation..  All rights reserved.
 */
 package mondrian.spi;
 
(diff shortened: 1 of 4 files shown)


Advent51:

Advent51:

Advent51:
There was wrong issue name in PR and commit message.

mondrian/src/main/java/mondrian/spi/Dialect.java
```diff
@@ -161,6 +161,18 @@ String quoteIdentifier(
...
+     * Encloses an identifier in quotation marks appropriate for the
+     * current SQL dialect. For example, in Oracle, where the identifiers
+     * are quoted using double-quotes,
+     * <code>quoteIdentifier("schema.name")</code> yields a string
+     * containing <code>"schema"."name"</code>.
+
+     * @param name Name to be quoted.
+     * @return Quoted identifier
+     */
+    String quoteIdentifierWithInnerDots(String name);
```
lucboudreau: I don't like adding a new method just for this one call. The logic that decides what to do with dots should be transparent to the caller and the dialect itself should hide that logic. I recommend removing this method and handling everything in Dialect.quoteIdentifier

Advent51: I cannot do that without changing the method signature.

Are you sure that I need to do that?

lucboudreau: Not sure I understand why this requires a change in signature. Is there a reason why ``quoteIdentifier(String val, StringBuilder buf)`` isn't ok?

Advent51: @lucboudreau yes, there is: we need to use the logic from `quoteIdentifierWithInnerDots` ONLY for schema name, and not for the others names (like table name or table alias).
To fully understand this, you should look at the https://github.com/pentaho/mondrian/pull/865/files#diff-ea35a8727d98201e0883f976f9cf6b34R808 and http://jira.pentaho.com/browse/MONDRIAN-2471

Advent51: @lucboudreau

lucboudreau: I understand now. So should we rename the new method to something clearer, like ``quoteSchemaIdentifier()`` ?

julianhyde: What is a user supposed to do if their schema name contains a dot?

Advent51:
closed as not a bug

PR #866, opened 2017-04-03
[MONDRIAN-2207] The segment cache index registry must use SchemaKey instead of only the checksum

Advent51:
- SchemaKey became public
- key getter was added into RolapSchema
- SegmentCacheIndexRegistry now use SchemaKey

### DIFF
--- a/mondrian/src/main/java/mondrian/rolap/agg/SegmentCacheManager.java
+++ b/mondrian/src/main/java/mondrian/rolap/agg/SegmentCacheManager.java
@@ -4,7 +4,7 @@
 // http://www.eclipse.org/legal/epl-v10.html.
 // You must accept the terms of that agreement to use this software.
 //
-// Copyright (c) 2002-2016 Pentaho Corporation.
+// Copyright (c) 2002-2017 Pentaho Corporation.
 // All Rights Reserved.
 */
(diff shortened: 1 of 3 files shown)


Advent51:

mondrian/src/main/java/mondrian/rolap/agg/SegmentCacheManager.java
```diff
@@ -1606,13 +1606,6 @@ public SegmentCacheIndex getIndex(RolapStar star) {
         private SegmentCacheIndex getIndex(
             SegmentHeader header)
         {
-            // First we check the indexes that already exist.
```
lucboudreau: Why remove this?

Advent51: Because `Checksum` was replaced by a `SchemaKey`, and we can retreive `SchemaKey` only from `star.getSchema().getKey()`.
If I will replace `header.schemaChecksum` with the `getStar(header).getSchema().getKey()` we will get the duplication of `getIndex(star)`

Advent51: @lucboudreau

PR #867, opened 2017-04-07
[PPP-3692] - Use of vulnerable component commons-beanutils: 1.8.0 - C…

YuryBY:
…VE-2014-0114

- remove redundant dependency commons-collections-core

@pamval, @mbatchelor, @graimundo could you please take a look?

no need exclusion more since there is no jar in kettle-core


YuryBY:
@graimundo, AFAIU broken tests don't relate to the changes in PR.
@lucboudreau, could you please take a look can it be wingman job configuration issue?
Thanks.

PR #868, opened 2017-04-10
[MONDRIAN-2424] Query defining set or member does not work

Advent51:
- lookupExp was changed - named set incorrect identification was fixed


Advent51:

PR #869, opened 2017-04-13
[SP-3410] Backport of MONDRIAN-2541 - Virtual cubes defined with only…

stanislau-strelchanka:
[SP-3410] Backport of MONDRIAN-2541 - Virtual cubes defined with only calc measures for one or more cubes can produce wrong results (7.0 Suite)


stanislau-strelchanka:
@lucboudreau @mkambol @kurtwalker @mchen-len-son please review

PR #870, opened 2017-04-14
[SP-3444] Backport of MONDRIAN-2424 - Query defining set or member does not work (7.0 Suite)

Advent51:


Advent51:

PR #879, opened 2017-04-14
[SP-3393] Backport of MONDRIAN-625 - The Descendants function will only return Non Empty member children, even if it is not specified inside of NON EMPTY (6.1 Suite)

Advent51:


Advent51:

PR #881, opened 2017-04-18
[PAD-159] Aggregate attributes are added to degenerate dimensions

stanislau-strelchanka:
- get rid of implicit MondrianDef.Hierarchy.relation set to cube.fact


stanislau-strelchanka:
PAD is generating schema.xml from MondrianDef. When RolapSchema and all it's dependent objects are created MondrianDef.Hierarchy is updated by RolapHierarchy constructor if there is no table specified in hierarchy. RolapHierarchy sets MondrianDef.Hierarchy.relation to fact table of a cube. So after that, when generating from modified schema relation element will be duplicated. Removed this implicit modification of MondrianDef.Hierarchy.relation.

@mkambol @lucboudreau @kurtwalker @mchen-len-son please review

PR #889, opened 2017-05-10
[COMP-5069] Removes dependencies on MySQL and PostgreSQL

lucboudreau:


lucboudreau:
@mkambol @kurtwalker For review.

PR #892, opened 2017-05-23
[SP-3553][PSW-252] - Missing ellipsis after Open and About

mchen-len-son:


mchen-len-son:
@mbatchelor @kurtwalker please take a look.
Thanks.

PR #898, opened 2017-06-02
[MONDRIAN-2253] - Pentaho analyzer is throwing error when multiple sh…

YuryBY:
…ared dimensions are used in a virtual cube - Vectorwise DB

 - reworked fix

### DIFF
--- a/mondrian/src/it/java/mondrian/test/DialectTest.java
+++ b/mondrian/src/it/java/mondrian/test/DialectTest.java
@@ -27,6 +27,7 @@ import java.util.*;
 import javax.sql.DataSource;
 
 import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.when;
 
 /**
  * Unit test which checks that {@link mondrian.spi.Dialect}
@@ -1660,12 +1661,16 @@ public class DialectTest extends TestCase {
     }
 
     public void testMondrian2253() throws SQLException {
+        String expected = "    1 ASC";
         // "1" is supposed to be a column number
-        String expected = "CASE WHEN 1 IS NULL THEN 1 ELSE 0 END, 1 ASC";
         String expr = "1";
         JdbcDialectImpl dialect = new VectorwiseDialect(getConnection());
+
         SqlQuery query = new SqlQuery(dialect, true);
-        query.addOrderBy(expr, null, true, true, true, true);
+        query.addOrderBy(
+            expr, null, true, false,
+            dialect.requiresUnionOrderByOrdinal(), true);
+
         assertTrue(query.toString().contains(expected));
     }
 }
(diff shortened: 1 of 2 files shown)


YuryBY:
@lucboudreau, @mbatchelor, could you please take a look?

lucboudreau:
Not sure if this is intentional, but the test went from testing both the alias and the ordering of null values to looking for '1 asc' only. Can we improve this test a little to make it more robust? It seems likely to me that '1 asc' could be found somewhere in the query and potentially cause false positives.

YuryBY:
@lucboudreau,  tabulation is essential here. Before fix we have:  
"order by
    CASE WHEN 1 IS NULL THEN 1 ELSE 0 END, 1 ASC"
and after fix: 
"order by
    1 ASC
"
So that test the is valid.
The reason way I did not use full compliance are os-specific ends of line (CRLF and LF). Otherwise it needs overhead code here to depend on os tests run.

lucboudreau:
Actually, I worry more about the change from ordering nulls to not ordering nulls. 

Have you tested the query on Vectorwise with null values present? Are they collated in the same order in both queries, before and after?

YuryBY:
@lucboudreau, I've leaved comment in jira: http://jira.pentaho.com/browse/MONDRIAN-2253?focusedCommentId=302476&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-302476.
Neither  "CASE WHEN "1" IS NULL THEN 1 ELSE 0 END, "1" ASC"
nor "CASE WHEN 1 IS NULL THEN 1 ELSE 0 END, 1 ASC" works for ingres vectorwise.
So that there is now possibility to check behavior "before". After the changes analyzer renders table with empty results: "-".
Did not investigate it more precisely.

YuryBY:
@lucboudreau, @mbatchelor, @mbatchelor, @pamval, could you please take a look?

YuryBY:
@lucboudreau, @mbatchelor, @mbatchelor, @pamval, any updates?

lucboudreau:
The method to order nulls has an argument which specifies how nulls are to be sorted. we need to make sure that your change to the dialect makes Vectorwise return the null values ordered correctly. That is very important, or wrong results will be returned. 

Please test on Vectorwise and compare with MySQL (just an example) to make sure that the null values are ordered the same across platforms.

YuryBY:
@lucboudreau, I tried following queries: https://github.com/pentaho/mondrian/compare/master...YuryBY:MONDRIAN-2253-tests?expand=1#diff-e4c97d55abf53b0277ee871148fe2444R1741all of them return the same results for MySQL and for Vectorwise. I'm not sure if it's enough for the validation. Could you please provide more query examples if it's needed? Thanks.

PR #899, opened 2017-06-05
[PSW-230] PSW Allows you to select Measures' datatypes that are not supported

Advent51:
- unsupported datatypes were removed

### DIFF
--- a/mondrian/src/main/java/mondrian/olap/Mondrian.xml
+++ b/mondrian/src/main/java/mondrian/olap/Mondrian.xml
@@ -966,7 +966,7 @@ Revision is $Id$
         <Attribute name="datatype" required="false">
             <Doc>
                 The datatype of this measure:
-                String, Numeric, Integer, Boolean, Date, Time or Timestamp.
+                String, Numeric, Integer.
 
                 <p/>The default datatype of a measure is
                 'Integer' if the measure's aggregator is 'Count',
@@ -975,10 +975,6 @@ Revision is $Id$
             <Value>String</Value>
             <Value>Numeric</Value>
             <Value>Integer</Value>
-            <Value>Boolean</Value>
-            <Value>Date</Value>
-            <Value>Time</Value>
-            <Value>Timestamp</Value>
         </Attribute>
         <Attribute name="formatString">
             <Doc>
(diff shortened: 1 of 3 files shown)


Advent51:
@lucboudreau @mkambol @mchen-len-son please review

mondrian/src/main/java/mondrian/olap/Mondrian.xml
```diff
@@ -966,7 +966,7 @@ Revision is $Id$
         <Attribute name="datatype" required="false">
             <Doc>
                 The datatype of this measure:
-                String, Numeric, Integer, Boolean, Date, Time or Timestamp.
+                String, Numeric, Integer.
```
mkambol: Boolean, Date, Time and Timestamp are intended to be supported as measure datatypes, if I'm not mistaken. 
http://mondrian.pentaho.com/documentation/schema.php#Measures

Tried a quick test with Boolean using count / distinct-count aggregators, seemed to work fine.

I think this bug should probably be to make PSW allow those data types.

Advent51: http://jira.pentaho.com/browse/MONDRIAN-707 saying that it is not.

mkambol: Thanks Andrei.  Those datatypes wouldn't make sense as sums, but seem useful to me with other aggregators-- e.g. count of dates, minimum time, etc.  

type="Boolean" works for me with a hand written schema.  Would be worth trying the other datatypes, too.

I looked around for the unit test that Will G. mentioned in MONDRIAN-707 to get some context on why he thought they should not be supported, but couldn't find it.  Did you?

Advent51: No, I didn't find it too.
So you think that PSW-230 is not a bug?

mkambol: Well, there's definitely a bug since we're throwing an exception when using a datatype that's specified as allowed.

Digging in a bit more- the only reason my test was passing is that I used the "type" attribute name instead of "datatype", so the boolean field was being treated as an integer.

I still see value in explicitly supporting those datatypes, but I guess that becomes more of an enhancement.  I'll go ahead and merge this one.

PR #907, opened 2017-07-07
[MONDRIAN-1398] The CurrentMember function should be disallowed when …

stanislau-strelchanka:
[MONDRIAN-1398] The CurrentMember function should be disallowed when the dimension is in a compound slicer

### DIFF
--- a/mondrian/src/it/java/mondrian/test/BasicQueryTest.java
+++ b/mondrian/src/it/java/mondrian/test/BasicQueryTest.java
@@ -8915,6 +8915,52 @@ public class BasicQueryTest extends FoodMartTestCase {
         testContext.assertQueryReturns(mdxWithBug, expectedResultInBugCube);
     }
 
+    public void testCurrentMemberWithCompoundSlicer() {
+        String mdx = ""
+                + "with\n"
+                + "member [Measures].[Gender Current Member] "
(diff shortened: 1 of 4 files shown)


stanislau-strelchanka:
@mchen-len-son @kurtwalker @lucboudreau @mkambol please review

lucboudreau:
@mkambol @stanislau-strelchanka I feel we could use a property to use the old behavior, although I'm hoping to have your thoughts on this as well. We have other fixes in place around disallowing certain calculations for exactitude's sake, and we've usually added a way to disable those checks. In this case, since the new behavior throws an exception, I'm inclined to add a legacy property. Thoughts?

mkambol:
IIRC, the old behavior would return the null member for .CurrentMember, which is almost certainly not what would be expected.  Analysis Services will error out if you attempt to do it.
Given that it would be unusual to even attempt it, and results will be wrong if you do, I'd say a property would just add confusion.

lucboudreau:
Other thing I wish I had caught earlier. We need to use the messages and errors bundles. We don't hard code error messages and exceptions into the code.

mkambol:
Good point.  @stanislau-strelchanka - can you do a CLEANUP commit for that?

mchen-len-son:
@mkambol, Stanislau is on vac - I will provide it

PR #908, opened 2017-07-10
[PPP-3566] Use of vulnerable component xerces:xercesImpl:2.9.1and 2.1…

mchen-len-son:
…1.0 -

CVE-2009-2625 and CVE-2013-4002
- removed xerces jars in order to switch to JRE bundled implementation
- replaced all explicit xerces calls with JAXP and DOM LS Specification use.
- checked source code of commons-dbcp and commons-pool of specified versions for explicit xerces usage - didn't find any - thus, excluded xerces for that dependencies.
- also excluded xerces for eigenbase-xom, since it is not "effectively" used there. XercesDOMParser is not used at code-generation, it is only used in tests.

### DIFF
--- a/assemblies/psw-ce/pom.xml
+++ b/assemblies/psw-ce/pom.xml
@@ -226,12 +226,6 @@
       <scope>runtime</scope>
     </dependency>
     <dependency>
-      <groupId>xerces</groupId>
-      <artifactId>xercesImpl</artifactId>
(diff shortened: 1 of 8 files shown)


stepanovdg:
https://github.com/pentaho/mondrian/pull/920 Tried to restart to get right wingman message.

PR #909, opened 2017-07-18
[MONDRIAN-2274]-Calculated members not defined on [Measures] which override context can suppress results in some limited cases

stepanovdg:


stepanovdg:
@mkambol @lucboudreau Could you please take a look?

mkambol:
This commit makes me nervous.  It leaks details of native eval and CJ optimizer logic into RolapResult, complicating the already complex process of axis evaluation.  Apart from code readability and understandability, I'm also worried about performance impact, and whether more "false positive" flags of conflicting calcs will result.

When we last chatted about 2274 on email, I mentioned:

> I created that case so we’d remember the limitation.  I’d push back on whether we should actually
> pick it up, though.  Analyzer will not generate MDX that would produce this issue, and afaik no 
> customer has reported hitting it.  Native evaluation logic is notoriously complex and nasty— I’d be
> cautious about attempting 2274 lightly.

I'm still of the same mind.  Don't touch native eval logic unless there's compelling reason to do so.  Do we have customers or the community that are reporting this issue?

PR #910, opened 2017-07-18
[CLEANUP] moved hard-coded error message to message bundle

mchen-len-son:


mchen-len-son:
@mkambol this is a cleanup for #907.
please take a look.

PR #916, opened 2017-07-31
[MONDRIAN-2408]: <Set> IIf(<Logical Expression>, <Set>, <Set>) confus…

TatsianaKasiankova:
…es mondrian into throwing ResultStyleException "Wanted ResultStyles: {LIST,MUTABLE_LIST} but got: VALUE"

GenericIterCalc that is used in Iif function for return type=SetType always returns ResultStyle.VALUE.

The fix is to get the correct ResultStyle for the function using the calculation of value parameters.
As Iif function takes the same types for second and third parameters (please see Iif function signature in http://mondrian.pentaho.com/documentation/mdx.php),
so we can use the any of them.

### DIFF
--- a/mondrian/src/it/java/mondrian/olap/fun/FunctionTest.java
+++ b/mondrian/src/it/java/mondrian/olap/fun/FunctionTest.java
@@ -5,7 +5,7 @@
 // You must accept the terms of that agreement to use this software.
 //
 // Copyright (C) 2003-2005 Julian Hyde
-// Copyright (C) 2005-2016 Pentaho and others
+// Copyright (C) 2005-2017 Pentaho and others
 // All Rights Reserved.
 */
 package mondrian.olap.fun;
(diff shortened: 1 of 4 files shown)


TatsianaKasiankova:
enericIterCalc that is used in Iif function for return type=SetType always returns ResultStyle.VALUE.

The fix is to get the correct ResultStyle for the function using the calculation of value parameters.
As Iif function takes the same types for second and third parameters (please see Iif function signature in http://mondrian.pentaho.com/documentation/mdx.php),
so we can use the any of them.
@mkambol , @lucboudreau please review.

TatsianaKasiankova:
I'd like to add a bit about *Still Broken Tests:* section in WINGAN checks.
We have a bunch of broken test on master for some time. I've built mondrian on master: please see http://10.177.176.33:8081/view/MONDRIAN/job/mondrian_314_maven/120/testReport/ and gotten the same failed tests.

mkambol:
This looks good.  Can you add a test to lock in the expected behavior?

PR #920, opened 2017-08-04
[PPP-3566] Use of vulnerable component xerces:xercesImpl:2.9.1and 2.11.0 -CVE-2009-2625 and CVE-2013-4002

stepanovdg:
- removed xerces jars in order to switch to JRE bundled implementation


tmcsantos:
Don't take the last report as granted, I had to kill the running test thread to continue the build since it was stuck for 2h.

lucboudreau:
Fixed by #920

PR #921, opened 2017-08-05
[ANALYZER-3488]: Unable to do case insensitive search with multi-byte characters in filter dialog and using the Matches operator

TatsianaKasiankova:
The problem:
The fix: https://github.com/pentaho/pentaho-analyzer/commit/021a081d26fc51e13753ae5210073f38d5519b2d was not completed in full.
It covers the DBs that are not supported native evalution of the Matches MDX function.
In this case mondrian uses Java regex engine to evaluate matching.
By default, case-insensitive matching assumes that only characters in the US-ASCII charset are being matched (Please see https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html#UNICODE_CASE).
So it was added (?u) flag to allow unicode-aware case folding. Please see details in https://github.com/pentaho/pentaho-analyzer/commit/021a081d26fc51e13753ae5210073f38d5519b2d.

But we have also BDs with implemented support of native evalution of the Matches MDX function (was introduced in MONDRIAN-944).
None of them does not supported (?u) flag.

So it was implemented additional fix:
1) Added the flag to inform if DB supports (?u) or not;
2)If DB dialect does not support (?) flag, then regular expression should be cleaned from this flag.
3)Added unit tests for all DB dialects with native support of regular expression.

### DIFF
--- a/mondrian/src/it/java/mondrian/spi/DialectUtilTest.java
+++ b/mondrian/src/it/java/mondrian/spi/DialectUtilTest.java
@@ -1,0 +1,37 @@
+/*
+// This software is subject to the terms of the Eclipse Public License v1.0
+// Agreement, available at the following URL:
+// http://www.eclipse.org/legal/epl-v10.html.
+// You must accept the terms of that agreement to use this software.
+//
+// Copyright (c) 2015-2017 Pentaho Corporation.
+// All rights reserved.
+ */
+package mondrian.spi;
+
+import junit.framework.TestCase;
+
+public class DialectUtilTest extends TestCase {
+
+  public void testCleanUnicodeAwareCaseFlag_InputNull() {
+    String inputExpression = null;
+    String cleaned = DialectUtil.cleanUnicodeAwareCaseFlag( inputExpression );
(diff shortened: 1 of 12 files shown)


TatsianaKasiankova:
@mkambol , please review the PR.

TatsianaKasiankova:
@mkambol , please review the PR.

mkambol:
Are there any databases that support the unicode aware flag?

TatsianaKasiankova:
I did not verified for ALL DBs. 
I've verified only for these DBs that returns true in allowsRegularExpressionInWhereClause() method in 
No one of them support the unicode aware flag.

mkambol:
Feels odd to me to be adding a property to the dialect for support of the unicode aware flag, but we're not aware of any databases that support it.

Also, if I'm reading your change right, you're simply stripping the (?u) out of the string, which I assume would mean that some databases would fail to do case insensitive matches on  unicode char fields.  Seems like we'd want to do something like wrap it in a LOWER(), or let it just return null (which it would do if the db didn't support regex in the where).

PR #923, opened 2017-08-08
[MONDRIAN-2561] Crossjoin results differs between sql and mondrian wh…

stanislau-strelchanka:
[MONDRIAN-2561] Crossjoin results differs between sql and mondrian when suppress zeros is on


stanislau-strelchanka:
@mbatchelor @lucboudreau @kurtwalker please review

PR #930, opened 2017-08-15
[MONDRIAN-1967] MondrianOlap4jMember.isVisible throws NullPointerExce…

stanislau-strelchanka:
[MONDRIAN-1967] MondrianOlap4jMember.isVisible throws NullPointerException

### DIFF
--- a/mondrian/src/it/java/mondrian/test/Olap4jTest.java
+++ b/mondrian/src/it/java/mondrian/test/Olap4jTest.java
@@ -4,7 +4,7 @@
 * http://www.eclipse.org/legal/epl-v10.html.
 * You must accept the terms of that agreement to use this software.
 *
-* Copyright (c) 2002-2013 Pentaho Corporation..  All rights reserved.
+* Copyright (c) 2002-2017 Pentaho Corporation..  All rights reserved.
 */
 
 package mondrian.test;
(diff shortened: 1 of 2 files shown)


stanislau-strelchanka:
@lucboudreau @mkambol @kurtwalker please review

kurtwalker:
when did these tests become broken?  Please fix them before merging

stanislau-strelchanka:
@kurtwalker same tests are failing here #928 where nothing is changed. So it looks like wingman issue.
Also I've checked them locally and using mondrian job (no errors)

stanislau-strelchanka:
@lucboudreau @mkambol @kurtwalker please review

stanislau-strelchanka:
@lucboudreau @mkambol @kurtwalker please review

kurtwalker:
Do you have an explanation for why wingman is giving different results than local and devci?  I don't want to merge unless we understand what's going on.

stanislau-strelchanka:
@kurtwalker We encountered similar problems before. I believe this might be environment issues caused by tests that are updating database and not always cleaning up after themselves ( in case of an error for example ).  Also I've seen errors where expected result is different from actual only by order of tuples and order wasn't defined. The last case could be the reason of tests that are passing / failing over time without no reason.

stanislau-strelchanka:
@lucboudreau @mkambol @kurtwalker please review

stanislau-strelchanka:
@lucboudreau @mkambol @kurtwalker please review

PR #940, opened 2017-08-28
[SP-3741]: Backport of MONDRIAN-2408 - <Set> IIf(<Logical Expression>…

TatsianaKasiankova:
…, <Set>, <Set>) confuses mondrian into throwing ResultStyleException "Wanted ResultStyles: {LIST,MUTABLE_LIST} but got: VALUE" (6.1 Suite)

This is the cherry-pick of the fix for [MONDRIAN-2408]: <Set> IIf(<Logical Expression>, <Set>, <Set>) confuses mondrian into throwing ResultStyleException "Wanted ResultStyles: {LIST,MUTABLE_LIST} but got: VALUE"

GenericIterCalc that is used in Iif function for return type=SetType always returns ResultStyle.VALUE.

The fix is to get the correct ResultStyle for the function using the calculation of value parameters.
As Iif function takes the same types for second and third parameters (please see Iif function signature in http://mondrian.pentaho.com/documentation/mdx.php),
so we can use the any of them.

Added unit tests to cover cases like in the issue.

### DIFF
--- a/src/main/mondrian/olap/fun/IifFunDef.java
+++ b/src/main/mondrian/olap/fun/IifFunDef.java
@@ -4,7 +4,7 @@
 * http://www.eclipse.org/legal/epl-v10.html.
 * You must accept the terms of that agreement to use this software.
 *
-* Copyright (c) 2002-2013 Pentaho Corporation..  All rights reserved.
+* Copyright (c) 2002-2017 Pentaho Corporation..  All rights reserved.
 */
 
 package mondrian.olap.fun;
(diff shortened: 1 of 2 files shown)


TatsianaKasiankova:
@mchen-len-son , @stepanovdg , @pamval please review

mchen-len-son:
@TatsianaKasiankova please move IifFunDefTest.java to appropriate folder

TatsianaKasiankova:
@mchen-len-son updated.

PR #942, opened 2017-08-29
[SP-3653]: Backport of ANALYZER-3488 - Unable to do case insensitive …

TatsianaKasiankova:
…search with multi-byte characters in filter dialog and using the Matches operator (7.1 Suite)

This is the cherry-pick of the fix for [ANALYZER-3488]: Unable to do case insensitive search with multi-byte characters in filter dialog and using the Matches operator
The problem:
The fix: https://github.com/pentaho/pentaho-analyzer/commit/021a081d26fc51e13753ae5210073f38d5519b2d was not completed in full.
It covers the DBs that are not supported native evalution of the Matches MDX function.
In this case mondrian uses Java regex engine to evaluate matching.
By default, case-insensitive matching assumes that only characters in the US-ASCII charset are being matched (Please see https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html#UNICODE_CASE).
So it was added (?u) flag to allow unicode-aware case folding for Java regex engine. Please see details in https://github.com/pentaho/pentaho-analyzer/commit/021a081d26fc51e13753ae5210073f38d5519b2d.

But we have also BDs with implemented support of native evalution of the Matches MDX function (was introduced in MONDRIAN-944).
None of them does not support (?u) flag. And moreove this flag doesn't need for them.
So the fix is just to clean up the expression from this (?u) flag.
Also added unit tests for all DB dialects with native support of regular expression.

### DIFF
--- a/mondrian/src/it/java/mondrian/spi/DialectUtilTest.java
+++ b/mondrian/src/it/java/mondrian/spi/DialectUtilTest.java
@@ -1,0 +1,37 @@
+/*
+// This software is subject to the terms of the Eclipse Public License v1.0
+// Agreement, available at the following URL:
+// http://www.eclipse.org/legal/epl-v10.html.
+// You must accept the terms of that agreement to use this software.
+//
+// Copyright (c) 2015-2017 Pentaho Corporation.
+// All rights reserved.
+ */
+package mondrian.spi;
+
+import junit.framework.TestCase;
+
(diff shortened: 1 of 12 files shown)


TatsianaKasiankova:
@mchen-len-son , @stepanovdg , @pamval

stepanovdg:
Tests are failing occasionally.

PR #944, opened 2017-08-29
[SP-3756][MONDRIAN-1967] MondrianOlap4jMember.isVisible throws NullPo…

mchen-len-son:
…interException


mchen-len-son:
@stepanovdg please take a look

PR #947, opened 2017-09-04
[MONDRIAN-2402]: Mondrian incorrectly resolves <String> Generate(<Set>, <String>, <String>) to <Set> Generate(<Set>, <Set>, <Symbol>)

TatsianaKasiankova:
The cause of the bug:
In mondrian schema a dimention can contain many hierarchies.
In mondrian schema it's allowed hierarchies without name attribute. In this case the hierarchy has the same name as its dimension.
When expression with a dimention is been parsing and resolving and it needs to get the hierarchy of this dimention, then it's returned null for the case when the dimention contains two hierarchy.
This is incorrect because we ignore the hierarchies with the same name as dimention name.
An this is the cause of the issue because the return type for the member Property function is taken exactly from this member resolved hierarchy levels.
Please see:
https://github.com/pentaho/mondrian/blob/master/mondrian/src/main/java/mondrian/olap/fun/PropertiesFunDef.java#L113
https://github.com/pentaho/mondrian/blob/master/mondrian/src/main/java/mondrian/olap/fun/PropertiesFunDef.java#L137-L159

The fix is to return the hierarchy of the dimention with the same name as dimention name, when the dimention contains more then one hierarchy.
Added unit tests.

### DIFF
--- a/mondrian/src/it/java/mondrian/olap/fun/FunctionTest.java
+++ b/mondrian/src/it/java/mondrian/olap/fun/FunctionTest.java
@@ -6970,6 +6970,28 @@ public class FunctionTest extends FoodMartTestCase {
         fail("should have timed out");
     }
 
+    //The test case for the issue: MONDRIAN-2402
+    public void testGenerateForStringMemberProperty() {
+      assertQueryReturns(
+          "WITH MEMBER [Store].[Lineage of Time] AS\n"
+          + " Generate(Ascendants([Time].CurrentMember), [Time].CurrentMember.Properties(\"MEMBER_CAPTION\"), \",\")\n"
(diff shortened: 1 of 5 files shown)


TatsianaKasiankova:
@mkambol , @lucboudreau  please review the PR.

TatsianaKasiankova:
@mkambol , @lucboudreau please review the PR. 
Concerning the tests are failed from time to time on WINGMAN I've reported the issue: http://jira.pentaho.com/browse/MONDRIAN-2601.
In short the problem is in different order of elements in Axis in Actual result vs hard-coded Expected.
All these tests are always passed  on my local env.

PR #964, opened 2017-10-12
[BACKLOG-18450] remove or fix version specific integration tests in Mondrian

jgminder:
@pentaho/2-1b please review. I could not duplicate the original issue but found some broken CREATE statements and fixed them while investigating. mediumtext column types are not available when the ENGINE type is MEMORY in MySQL.


tmcsantos:
You may not having the issue because of this https://github.com/pentaho/mondrian/commit/8eb44dd0f4c522067d30cf4ca3877df502a5f3dd

try reverting that and see if it still fails.

PR #967, opened 2017-10-18
[PDI-16624]-Big Number fields are not correctly aggregated in DET. Added a new PdiDataServiceDialect

ArtsiomYurhevich:
Divided into 2 commits in order not to confuse pr changes with checkstyle

### DIFF
--- a/mondrian/src/main/java/mondrian/olap/Mondrian.xml
+++ b/mondrian/src/main/java/mondrian/olap/Mondrian.xml
@@ -1367,6 +1367,7 @@ Revision is $Id$
                 <li>luciddb</li>
                 <li>vertica</li>
                 <li>neoview</li>
+                <li>pdi</li>
                 </ul>
             </Doc>
         </Attribute>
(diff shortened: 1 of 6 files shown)


ArtsiomYurhevich:
@mkambol , @mbatchelor Could you please review the reworked solution. Thanks

mkambol:
I think this looks good.  
For style, though-- I believe we're still using the non-pentaho "checkFile" conventions for Mondrian.  @lucboudreau, is that true?  Also, @lucboudreau - anything else to consider when creating a new dialect then what @ArtsiomYurhevich has done?

lucboudreau:
Please use Mondrian's coding standard with the bin/checkStyle utility.

You will also need to add the dialect name in mondrian.xml and mondrian_sw.xml.

ArtsiomYurhevich:
@mkambol , @lucboudreau Done

lucboudreau:
Looks good.

Side question, is there a jira case to fix those tests?

PR #976, opened 2017-11-10
[MONDRIAN-2597]: readTuples and cardinality queries sent twice to the…

TatsianaKasiankova:
… database when using Virtual Cube (Not cached)

Fixed the twice sending of readTuples query to the DB.

### DIFF
--- a/mondrian/src/it/java/mondrian/rolap/NativeEvalVirtualCubeTest.java
+++ b/mondrian/src/it/java/mondrian/rolap/NativeEvalVirtualCubeTest.java
@@ -383,6 +383,114 @@ public class NativeEvalVirtualCubeTest extends BatchTestCase {
     assertQuerySqlOrNot(
         tc, mdx, new SqlPattern[]{ mysqlPattern }, true, false, false);
   }
+
+  /**
+   * Testcase for bug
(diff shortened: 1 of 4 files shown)


TatsianaKasiankova:
@lucboudreau , @mkambol please review the fix for the part related with duplicating `readTuples`.

mkambol:
@TatsianaKasiankova -  it feels to me like the list of base cubes should be something you should be able to learn from the Virtual cube itself.  That is, instead of spreading the logic for getting/setting the list of base cubes across mutable fields in the Query and special purpose logic in SqlContextConstraint, is there a way we could include it in RolapCube?  RolapCube already has a .isVirtual method.  Seems like it would be reasonable to also have a .baseCubes() method.  That would also be a consistent way of getting the info, regardless of whether native eval is in use or not (SqlContextConstraint won't be used with non-native).

@kurtwalker - I know you looked at this earlier, would be curious if you have an opinion.

mkambol:
Oh, also-  could you add one or more tests?

kurtwalker:
I agree, it could go on RolapCube and be a lazy-loaded getter.  In the case where it's not virtual, could just return this;

TatsianaKasiankova:
@lucboudreau , @mkambol , @kurtwalker please review the fix.

PR #977, opened 2017-11-10
[PDI-15981]-Error when using Model View with Measure name having Japa…

ArtsiomYurhevich:
…nese characters


ArtsiomYurhevich:
@mbatchelor , @bennychow , @lucboudreau Could you please take a look?

PR #981, opened 2017-11-30
[SP-3922]: Backport of MONDRIAN-1073 - Two cubes operating on same fact table gives wrong WHERE clause (8.0 Suite)

TatsianaKasiankova:
This is the cherry-pick of the fix for [MONDRIAN-1073]: Two cubes operating on same fact table gives wrong WHERE clause

There are fix, unit tests and the integration test.


TatsianaKasiankova:
Wrong `into `branch.

PR #982, opened 2017-11-30
[SP-3921]: Backport of MONDRIAN-1073 - Two cubes operating on same fact table gives wrong WHERE clause (7.1 Suite)

TatsianaKasiankova:
This is the cherry-pick of the fix for [MONDRIAN-1073]: Two cubes operating on same fact table gives wrong WHERE clause

There are fix, unit tests and the integration test.


TatsianaKasiankova:
@mchen-len-son , @stepanovdg please review

PR #997, opened 2017-12-26
[SP-3956]: Backport of MONDRIAN-2603 - Measures completely disappear from a second cube defined in the same schema when measures are constrained using Mondrian role (7.1 Suite)

TatsianaKasiankova:
This is the cherry-pick of the fix for [MONDRIAN-2603]: Measures completely disappear from a second cube defined in the same schema when measures are constrained using Mondrian role (#994)


TatsianaKasiankova:

TatsianaKasiankova:
The Broken Tests above are failed from time to time on Wingman, although the same tests are passed successfully on the other env (please see reported issue for this case http://jira.pentaho.com/browse/BACKLOG-18779).

PR #998, opened 2017-12-26
[SP-3957]: Backport of MONDRIAN-2603 - Measures completely disappear from a second cube defined in the same schema when measures are constrained using Mondrian role (8.0 Suite)

TatsianaKasiankova:
This is the cherry-pick of the fix for [MONDRIAN-2603]: Measures completely disappear from a second cube defined in the same schema when measures are constrained using Mondrian role (#994)


TatsianaKasiankova:

PR #1001, opened 2018-01-11
[MONDRIAN-2608]: Error with cached data when ordinalColumn is defined and parameter mondrian.native.ExpandNonNative is set to true

TatsianaKasiankova:
There are fix, IT test and check-style fixes.

### DIFF
--- a/mondrian/src/main/java/mondrian/rolap/RolapMemberBase.java
+++ b/mondrian/src/main/java/mondrian/rolap/RolapMemberBase.java
@@ -513,9 +513,17 @@ public class RolapMemberBase
 
     void setOrdinal(int ordinal) {
         if (this.ordinal == -1) {
-            this.ordinal = ordinal;
+          this.ordinal = ordinal;
         }
     }
+
+    protected void setOrdinal(int ordinal, boolean forced) {
+      if (forced) {
+          this.ordinal = ordinal;
+      } else {
+        setOrdinal(ordinal);
+      }
+  }
 
     void setOrderKey(Comparable orderKey) {
         this.orderKey = orderKey;
--- a/mondrian/src/it/java/mondrian/test/BasicQueryTest.java
+++ b/mondrian/src/it/java/mondrian/test/BasicQueryTest.java
@@ -6390,6 +6390,153 @@ public class BasicQueryTest extends FoodMartTestCase {
             + "Row #2: 30,114\n");
     }
 
+    /**
+     * Tests that members are returned in correct order
+     *  when ordinalColumn is defined for the level
+     *  and members are cached partially already by previous MDX; bug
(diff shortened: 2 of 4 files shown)


TatsianaKasiankova:
@lucboudreau , @kurtwalker, @mkambol please review.
The problem only has place when `mondrian.rolap.compareSiblingsByOrderKey=false` and `ordinalColumn` is defined.
In this case during tuple reading and base members creation no `orderKey` will be set up, but only `ordinal` (please see https://github.com/pentaho/mondrian/blob/master/mondrian/src/main/java/mondrian/rolap/SqlMemberSource.java#L1089-L1093)
In further we wrap every base member into RolapCubeMember (https://github.com/pentaho/mondrian/blob/master/mondrian/src/main/java/mondrian/rolap/RolapCubeHierarchy.java#L774-L777) and return this list of members.
But during wrapping we use already cached RolapCubeMember objects (e.g. during any previous MDX running) and they have base members with `ordinal` values from previous MDX running (see https://github.com/pentaho/mondrian/blob/master/mondrian/src/main/java/mondrian/rolap/RolapCubeHierarchy.java#L836-L840).
Then when we are sorting such members list in  the method `mondrian.olap.fun.FunUtil.hierarchizeMemberList(List<Member>, boolean)`, we using `ordinal` values of  base members in RolapCubeMember and get incorrect members order as result (see https://github.com/pentaho/mondrian/blob/master/mondrian/src/main/java/mondrian/olap/fun/FunUtil.java#L1975-L1986).

mondrian/src/main/java/mondrian/rolap/RolapMemberBase.java
```diff
@@ -513,10 +513,16 @@ public Comparable getOrderKey() {
 
     void setOrdinal(int ordinal) {
         if (this.ordinal == -1) {
-            this.ordinal = ordinal;
+          setOrdinal(ordinal, true);
         }
     }
 
+    protected void setOrdinal(int ordinal, boolean forced) {
+      if (forced) {
```
kurtwalker: This seems weird.  If the caller knows enough to pass forced or not, wouldn't they just be able to not call the method when forced is false?

TatsianaKasiankova: I need the method just to set up ordinal.
In this class we already have the method `void setOrdinal(int ordinal)` with condition:
  `if (this.ordinal == -1)`
This method is using in many places and I don't want to delete this condition because this can affect already existing logic.
That is why I've just overloaded `setOrdinal` method with bool parameter.

TatsianaKasiankova: Changed a bit to not miss `false` case.

PR #1002, opened 2018-01-12
[PDI-16761] In Model View, integer type numerical value more than 17 digits is rounded off

AliaksandrShuhayeu:
[PDI-16761] In Model View, integer type numerical value more than 17 digits is rounded off

added correct processing BigDecimal without losing precision


AliaksandrShuhayeu:
@kurtwalker Could you please review and merge it?

PR #1013, opened 2018-01-24
[SP-4038]: Backport of MONDRIAN-2608 - Error with cached data when ordinalColumn is defined and parameter mondrian.native.ExpandNonNative is set to true (7.1 Suite)

TatsianaKasiankova:
This is the cherry-pick of the fix for [MONDRIAN-2608]: Error with cached data when ordinalColumn is defined and parameter mondrian.native.ExpandNonNative is set to true (#1001)
There are fix, IT test and check-style fixes.


TatsianaKasiankova:

PR #1019, opened 2018-02-01
[MONDRIAN-2604]: java.lang.ArrayIndexOutOfBoundsException while applying Filters

TatsianaKasiankova:
There are fix, unit tests and a couple of checkstyle fixes.

### DIFF
--- a/mondrian/src/it/java/mondrian/rolap/RolapStarTest.java
+++ b/mondrian/src/it/java/mondrian/rolap/RolapStarTest.java
@@ -4,7 +4,7 @@
 * http://www.eclipse.org/legal/epl-v10.html.
 * You must accept the terms of that agreement to use this software.
 *
-* Copyright (c) 2002-2017 Hitachi Vantara..  All rights reserved.
+* Copyright (c) 2002-2018 Hitachi Vantara..  All rights reserved.
 */
 
 package mondrian.rolap;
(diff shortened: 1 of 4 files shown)


TatsianaKasiankova:
@lucboudreau , @mkambol , @kurtwalker please review.
This problem only has place for the DBs that support Grouping Sets and have the columns with the same names in different tables.
In this case on creation a new `GroupingSetsList` object we lose the columns with the same names because of issue in ColumnComparator during creation of roll-up columns set:
https://github.com/pentaho/mondrian/blob/master/mondrian/src/main/java/mondrian/rolap/agg/GroupingSetsList.java#L63
https://github.com/pentaho/mondrian/blob/master/mondrian/src/main/java/mondrian/rolap/agg/GroupingSetsList.java#L153-L154

This leads to the situation when columns of two different grouping sets will have the same  `groupingColumnsBitKey` (please see https://github.com/pentaho/mondrian/blob/master/mondrian/src/main/java/mondrian/rolap/agg/GroupingSetsList.java#L118-L120).
And then on data loading in `SegmentLoader` we get wrong cohort object and as result - the reported issue.

So the fix is to consider two columns equal only if their names are equal and they are from the same table.

TatsianaKasiankova:
@mkambol , @kurtwalker please review.

PR #1024, opened 2018-02-16
[BACKLOG-21393] Adds date predicate dialect methods for Google BigQuery.

lucboudreau:


lucboudreau:
@pentaho/ackbar @mkambol @kurtwalker @bennychow For review. Adds dialect methods for dates, time and timestamps. This is required because BQ does not allow comparisons of Date/Time with strings. We must cast the right side of the predicate into the proper type.

https://cloud.google.com/bigquery/docs/reference/legacy-sql#castingfunctions

PR #1026, opened 2018-02-26
[SP-4122][MONDRIAN-2577]-Missing table name when using filter find with aggregate tables

stepanovdg:


stepanovdg:
@mchen-len-son Could you please take a look?

PR #1031, opened 2018-03-08
[MONDRIAN-2615] Type of level member properties is passed as null - i…

pentaho-lmartins:
…gnores schema

Please see JIRA case in a few minutes with notes and questions.
For review only, please do not merge. Missing test cases and code style formatting is not current.

@duarteteixeira @pamval @lucboudreau @mkambol

### DIFF
--- a/mondrian/src/main/java/mondrian/olap/Property.java
+++ b/mondrian/src/main/java/mondrian/olap/Property.java
@@ -5,14 +5,14 @@
 // You must accept the terms of that agreement to use this software.
 //
 // Copyright (C) 2001-2005 Julian Hyde
-// Copyright (C) 2005-2017 Hitachi Vantara and others
+// Copyright (C) 2005-2018 Hitachi Vantara and others
 // All Rights Reserved.
 //
 // jhyde, 12 September, 2002
 */
-
 package mondrian.olap;
 
+import mondrian.rolap.SqlStatement;
 import mondrian.spi.PropertyFormatter;
 
 import java.util.*;
@@ -61,13 +61,31 @@ import java.util.*;
 public class Property extends EnumeratedValues.BasicValue {
 
     public enum Datatype {
-        TYPE_STRING,
-        TYPE_NUMERIC,
-        TYPE_BOOLEAN,
-        TYPE_DATE,
-        TYPE_TIME,
-        TYPE_TIMESTAMP,
-        TYPE_OTHER
+        TYPE_STRING(null),
+        TYPE_NUMERIC(null),
+        TYPE_INTEGER(SqlStatement.Type.INT),
+        TYPE_LONG(SqlStatement.Type.LONG),
+        TYPE_BOOLEAN(null),
+        TYPE_DATE(null),
+        TYPE_TIME(null),
+        TYPE_TIMESTAMP(null),
+        TYPE_OTHER(null);
+
+        private SqlStatement.Type type;
+
+        Datatype(SqlStatement.Type type) {
+            this.type = type;
+        }
+
+        public SqlStatement.Type getInternalType() {
+            return type;
+        }
+
+        public boolean isNumeric() {
+            return this == TYPE_NUMERIC
+              || this == TYPE_INTEGER
+              || this == TYPE_LONG;
+        }
     }
 
     /**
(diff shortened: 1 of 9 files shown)


mondrian/src/main/java/mondrian/olap/Property.java
```diff
@@ -755,6 +757,16 @@ public Datatype getType() {
         return type;
     }
 
+    public SqlStatement.Type getSqlType() {
```
lucboudreau: This method is not needed. If the schema overrides the type, we use it, if not we don't. No guessing.

mondrian/src/main/java/mondrian/rolap/SqlMemberSource.java
```diff
@@ -682,7 +682,7 @@ public Object getMemberCacheLock() {
                 hierarchy.addToFrom(sqlQuery, exp);
             }
             final String s = exp.getExpression(sqlQuery);
-            String alias = sqlQuery.addSelect(s, null);
+            String alias = sqlQuery.addSelect(s, property.getSqlType());
```
lucboudreau: See my comment above. Just pass the type, null or whatever else was defined in the schema.

mondrian/src/main/java/mondrian/rolap/SqlTupleReader.java
```diff
@@ -1515,7 +1515,7 @@ protected void addLevelMemberSql(
                 } else {
                     propSql = property.getExp().getExpression(sqlQuery);
                 }
-                final String propAlias = sqlQuery.addSelect(propSql, null);
+                final String propAlias = sqlQuery.addSelect(propSql, property.getSqlType());
```
lucboudreau: Same here.

pentaho-lmartins:
@lucboudreau @mkambol Updated PR with suggestions in JIRA case. Please review.
Now working on test cases to cover the changes introduced in this PR. Will update soon.

lucboudreau:
OK so first off, we can't change FoodMart like this. Please leave the columns as Numeric.

Second, there are a number of test failures. Not sure if they are caused by the schema change, but once you revert it, we'll know.

Lastly, there are no tests for the new long type. You need to create tests for those specifically.

mondrian/src/main/java/mondrian/olap4j/MondrianOlap4jProperty.java
```diff
@@ -94,7 +94,11 @@ public Datatype getDatatype() {
         case TYPE_BOOLEAN:
             return Datatype.BOOLEAN;
         case TYPE_NUMERIC:
-            return Datatype.UNSIGNED_INTEGER;
```
pentaho-lmartins: This is what makes the test cases fail and why I changed the FoodMart.xml file.
Just reverted FoodMart.xml.

pentaho-lmartins:
@lucboudreau I'm still working on the new test cases for the Long property.
However, I think some tests will need to be changed because of my comment above.

mondrian/src/main/java/mondrian/rolap/RolapLevel.java
```diff
@@ -412,7 +412,9 @@ private Property lookupProperty(List<Property> list, String propertyName) {
         } else if (type.equals("Numeric")) {
             return Property.Datatype.TYPE_NUMERIC;
         } else if (type.equals("Integer")) {
-            return Property.Datatype.TYPE_NUMERIC;
```
lucboudreau: We should probably add a parameter for this. I think the new behavior is better, but some people may have been relying on the old Integer->Numeric mapping. Side effects of this may be, for example, that reports which displayed numbers in the 0.00 format would change to 0. People using a custom member formatter may also encounter class cast exceptions.

You can create a new Mondrian property by adding it to

mondrian/src/main/java/mondrian/olap4j/MondrianOlap4jProperty.java
lucboudreau: Yeah I noticed this and thought this looked wrong. I even included it in the changes I suggested on the Jira. Numeric to unsigned integer is wrong, because you loose the decimals and the sign. The tests will need to change.

lucboudreau:
There are still failures caused by this pull request.

cravobranco:
A new PR was created to cover the required changes.
Please refer to: https://github.com/pentaho/mondrian/pull/1062

PR #1042, opened 2018-04-05
[BACKLOG-22417] Adds a sequential suffix to drillthrough measures if …

lucboudreau:
…the alias is already taken.


mondrian/src/main/java/mondrian/rolap/agg/DrillThroughQuerySpec.java
```diff
@@ -139,9 +139,20 @@ public int getMeasureCount() {
...
-            : columnNames.get(columnNames.size() - 1);
+        String alias =
+            request.getDrillThroughMeasures().size() > 0
+                ? request.getDrillThroughMeasures().get(i).getName()
+                : columnNames.get(columnNames.size() - 1);
+        int j = 0;
+        String maybe = alias;
+        while (columnNames.contains(maybe)
+            // This is dumb, but the last column is the measure.
+            && columnNames.indexOf(maybe) != (columnNames.size() - 1))
```
bennychow: I'm not following this logic... can you describe in what situations were we generating duplicate aliases?   It seems like the makeAlias method should have made the aliases distinct.

lucboudreau: The names always contain the default measure in the last position. That's because it's how XMLA works with default columns on drillthroughs. The logic goes 1) add all cell region keys 2) add default measure 3) if extended context is enabled, add unrelated dim keys; all in that order.

The code puts all the alias in one list sequentially and you can't tell which is the measure and which is not other than their position.

I tried to split it initially and broke a lot of stuff related to XMLA drillthrough expectations. I figured I'd leave it there for now until we fix this correctly, or the specs change... albeit unlikely.

PR #1044, opened 2018-04-12
[BACKLOG-22838] Fixes a false positive on alias conflict detection. I…

lucboudreau:
…f a column is not in the selct clause, it cannot be aliassed.


lucboudreau:
Please do not merge right away. I saw a weird problem earlier related to a leaked connection. I made some changes to the test to be sure it was the cause of it, but it only appeared on Wingman. Waiting on wingman results.

lucboudreau:
@bennychow For review.

PR #1055, opened 2018-05-31
Introduce Locus.isEmpty to prevent exceptions

blacelle:
Hello. We observe queries execution time from 10sec to 2 seconds by not trying to peek from Lokus when it is empty

### DIFF
--- a/mondrian/src/main/java/mondrian/server/Execution.java
+++ b/mondrian/src/main/java/mondrian/server/Execution.java
@@ -96,7 +96,11 @@ public class Execution {
     {
         Execution parentExec;
         try {
-            parentExec = Locus.peek().execution;
+            if (Locus.isEmpty()) {
+                parentExec = null;
+            } else {
+                parentExec = Locus.peek().execution;
+            }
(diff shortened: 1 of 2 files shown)


lucboudreau:
We cannot accept this change. A Locus is required for all executions. We need to be able to channel logs and events back to the parent mondrian instance, and the Locus carries this information.

If you're running into an issue where the locus is not set, I'm guessing that you are accessing the schema API directly. I suggest using withLocus() on the schema reader object. Does this correspond to what you are doing? Can you explain where and how you are seeing the empty Locus?

blacelle:
In my case (a simple drilldown query from Excel), I indeed encountered an empty collection of Locus. Not manipulating the schema API directly

lucboudreau:
Can you share a stack trace with us please? It would be a bug if this happens from inside of Mondrian.

lucboudreau:
Closing for now. Locus should never be empty. Let's discuss this in a Jira ticket instead.

https://jira.pentaho.com

PR #1056, opened 2018-05-31
Enable caching member properties even if data-cache is disabled

blacelle:
We observed MDX execution timings dropping from 60s to 10s by caching member details, even if data-caching is enabled.

### DIFF
--- a/mondrian/src/main/java/mondrian/olap/MondrianProperties.xml
+++ b/mondrian/src/main/java/mondrian/olap/MondrianProperties.xml
@@ -613,6 +613,19 @@ query to the next (the cache is cleared after each query).
         <Default>false</Default>
     </PropertyDefinition>
     <PropertyDefinition>
+        <Name>ForceMemberCaching</Name>
+        <Path>mondrian.rolap.star.forceMemberCaching</Path>
+        <Category>Caching</Category>
+        <Description>
+Boolean property that controls whether Axis Members
+properties (especially member children) should be cached.
+If true, member children will be cached even if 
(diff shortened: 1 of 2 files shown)


lucboudreau:
We cannot accept this pull request as is. There are a number of other places where members are cached. We need all of them to check for this property. We also need some tests demonstrating the behavior on and off.

I'm not a fan of the solution either. Properties that depend on each other are a sign of an architectural problem. It seems to fix a single corner case. What's the issue exactly? Why disable the cache globally and then enable piece parts? What part of your schema is changing frequently and needs the cache to be disabled? Maybe it's better to disable only that changing part, and not the other way around.

blacelle:
> There are a number of other places where members are cached

This seemed the only place where SmartMemberReader is instantiated. Where are the other places where members are cached?

> Properties that depend on each other are a sign of an architectural problem

Indeed. There is an existing option to enable caches in various form. This new switch proposes to activate a sub-set of caches options. It seemed simpler both to code and to configure this way, than having a switch to deactivate a sub-set of caches options.

The issue is given a large pivot table (e.g. 10.000 rows, due to a drill-down along a medium cardinality hierarchy), Mondrian ends requesting thousands of times a single SQL query to underlying SQL engine .

We disabled the cache to prevent caching aggregates values (as the underlying SQL tables are updated). However, the properties of existing members are stable.

lucboudreau:
Closing this PR for now. This requires proper architectural review. We can't just add arbitrary properties to disable this-and-that without a proper review of how those things interact. Please create a Jira case for further discussion.

https://jira.pentaho.com

PR #1061, opened 2018-06-18
[MONDRIAN-2634] Expose Level datatype through olap4j

bennychow:

PR #1063, opened 2018-06-19
[MONDRIAN-2634] Makes the level method a default of XmlaExtra to prev…

lucboudreau:
…ent backwards compat issues with existing impls.


lucboudreau:
This is to fix a problem introduced in #1061

PR #1074, opened 2018-10-02
[MONDRIAN-2342] - Xmla response does not contain properties in tuple …

cravobranco:
…section

* Whenever an element does not exist, print the affected property in tuple section of xmla response.

@pamval , @pentaho-lmartins , @mbatchelor , @lucboudreau , @mkambol , @kurtwalker Please review and merge if it's all ok.


cravobranco:
@mkambol , it happened because the query didn't return valid values to fill it (e.g. no entries when year = 1998).
Now I updated the PR with a query which returns some valid results.

PR #1082, opened 2018-11-08
[MONDRIAN-2633] - Regression: Aggregate Tables not hit when Using a D…

cravobranco:
…imension Filter

* Fixed a side effect introduced in MONDRIAN-2285.
* Improved mechanism in aggregate table: a constraint will not be generated because name column is not the same as key column.

### DIFF
--- a/mondrian/src/main/java/mondrian/rolap/SqlConstraintUtils.java
+++ b/mondrian/src/main/java/mondrian/rolap/SqlConstraintUtils.java
@@ -6,7 +6,7 @@
 //
 // Copyright (C) 2004-2005 TONBELLER AG
 // Copyright (C) 2005-2005 Julian Hyde
-// Copyright (C) 2005-2017 Hitachi Vantara and others
+// Copyright (C) 2005-2018 Hitachi Vantara and others
 // All Rights Reserved.
 */
 package mondrian.rolap;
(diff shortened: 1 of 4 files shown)


cravobranco:
@lucboudreau , @kurtwalker , @mkambol , @pentaho-lmartins please review and merge if it's all ok. Thanks!

mondrian/src/main/java/mondrian/rolap/SqlConstraintUtils.java
```diff
@@ -1674,6 +1674,12 @@ public static String constrainLevel(
                 // as the key column
                 int bitPos = column.getBitPosition();
                 AggStar.Table.Column aggColumn = aggStar.lookupColumn(bitPos);
+
+                if(aggColumn == null){
+                    LOG.warn("Aggregate table: A constraint will not be generated because name column is not the same as key column");
```
lucboudreau: This message should be externalized in ``mondrianresource.xml``

cravobranco: Agree, thanks!

cravobranco:
@lucboudreau , thanks for your comments. PR is now updated so it's available again for review.

PR #1083, opened 2018-11-14
[MONDRIAN-2640] - SqlMemberSource.getMemberChildren queries fired dow…

cravobranco:
…n to the lowest level when using role with topLevel thus causing performance issues

* Adjusted implementation of MONDRIAN-2440 in order to avoid a side effect.


cravobranco:
@lucboudreau , @kurtwalker , @mkambol , @pentaho-lmartins Please review and merge if it's all ok. Thanks!

PR #1106, opened 2019-02-26
[MONDRIAN-2426] Performance with subtotals/totals is significantly wo…

bennychow:
…rse than without.  Add function to be used by Analyzer.

### DIFF
--- a/mondrian/src/it/java/mondrian/olap/fun/CachedExistsTest.java
+++ b/mondrian/src/it/java/mondrian/olap/fun/CachedExistsTest.java
@@ -1,0 +1,262 @@
+/*
+// This software is subject to the terms of the Eclipse Public License v1.0
+// Agreement, available at the following URL:
+// http://www.eclipse.org/legal/epl-v10.html.
+// You must accept the terms of that agreement to use this software.
+//
+// Copyright (c) 2002-2019 Hitachi Vantara.  All rights reserved.
+*/
+package mondrian.olap.fun;
+
+import mondrian.test.FoodMartTestCase;
+
+/**
+ * Tests the CachedExists function.
+ *
+ * @author Benny Chow
+ */
+public class CachedExistsTest extends FoodMartTestCase {
+
+  public CachedExistsTest() {
+    super();
+  }
+
+  public CachedExistsTest( String name ) {
+    super( name );
+  }
+
+  public void testEducationLevelSubtotals() {
+    String query =
+        "WITH "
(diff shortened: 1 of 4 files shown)


bennychow:
Please see this comment for background on this function:

https://jira.pentaho.com/browse/MONDRIAN-2426?focusedCommentId=362760&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-362760

bennychow:
@mkambol @lucboudreau @kurtwalker Could someone help review this custom function I added for improving Analyzer performance?   Thanks

bennychow:
Here are the Analyzer changes that use this new function:

https://github.com/pentaho/pentaho-analyzer/pull/1935

lucboudreau:
I'm not sure the keys generated have enough information to be unique. A virtual cube using the same level name would require different subtotals but hit the same cache key, no?

bennychow:
@lucboudreau I don't think whether the cube is virtual or not would matter.  The generated keys (i.e. tuples of member unique names) are used to look up items from the input non-dynamic set.  Whether this is used in a virtual cube context or not should not matter.  There is no nonempty logic in the function.

lucboudreau:
@bennychow I think it may cause issues if a report is ran on the base cube first, then on a virtual cube with the same D/H/L names afterwards. A test would prove me wrong.

bennychow:
@lucboudreau This cache is query specific.   That’s why I put it into the Query object. Two different MDX statements would not share the same cache.  BTW, how long does the Query object live around for?  It is freed for garbage collection right after the query is done?

lucboudreau:
@bennychow We don't force garbage collection. It'll get picked up whenever the JVM needs more space.

lucboudreau:
My only last question is: should this be in the core function def for Exists? Is there a reason we wouldn't want this as the default behavior?

bennychow:
CachedExists assumes the first set is non-dynamic and requires the caller to name this set and pass that same name in on subsequent calls to benefit from the cache.  There’s also overhead for building this cache which may be unnecessary if Exists is only called once.  

I think this CachedExists is very specific to Analyzer and hard to generalize to Exists.

lucboudreau:
I can't remember if we have more documentation to update. I'm on my phone and can't check, and I'll take a look tomorrow, but there may be a MDX document to keep updated in the git project somewhere. 

And yeah. The first set turning static is a blocker for the general case Exists.

lucboudreau:
Hey @bennychow sorry about the delay. The MDX functions need to be documented here too.

https://github.com/pentaho/mondrian/blob/master/mondrian/src/site/resources/doc/mdx.html

bennychow:
@lucboudreau Hi Luc, I added the description to the function list HTML.  If this looks good, can you merge this PR?  Thanks

PR #1108, opened 2019-02-27
[MONDRIAN-2652] Bad SQL on readTuples query from Virtual Cube using M…

LeonardoCoelho71950:
…ondrian Roles

@pentaho/tatooine @pentaho-lmartins 
@mkambol @lucboudreau

### DIFF
--- a/mondrian/src/it/java/mondrian/test/SteelWheelsSchemaTest.java
+++ b/mondrian/src/it/java/mondrian/test/SteelWheelsSchemaTest.java
@@ -4,7 +4,7 @@
 // http://www.eclipse.org/legal/epl-v10.html.
 // You must accept the terms of that agreement to use this software.
 //
-// Copyright (c) 2002-2017 Hitachi Vantara..  All rights reserved.
+// Copyright (c) 2002-2019 Hitachi Vantara..  All rights reserved.
 */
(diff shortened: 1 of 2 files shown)


LeonardoCoelho71950:
> Looks good. Can you add a test?

Sure :)

mondrian/src/it/java/mondrian/test/SteelWheelsSchemaTest.java
```diff
@@ -2592,6 +2592,144 @@ public void testMondrian2411_3() throws Exception {
             + "{[Customer_DimUsage.Customers Hierarchy].[1 rue Alsace-Lorraine].[Roulet]}\n"
             + "Row #0: 1,701.95\n");
     }
+
+    /**
+     * this tests the fix for
+     * http://jira.pentaho.com/browse/Mondrian-2652
+     */
+    public void testMondrian2652() {
```
lucboudreau: Tests based on SteelWheels need to check if the test can run before it does. See the first few lines in test "testMeasures" above for an example.

LeonardoCoelho71950: Great, that was the problem!

PR #1109, opened 2019-03-05
[MONDRIAN-2655] Add AggregateScanSchema and AggregateScanCatalog conn…

cdeptula:
…ection properties to limit the catalogs and schemas that are scanned for tables when establishing the database connection

### DIFF
--- a/mondrian/src/it/java/mondrian/rolap/aggmatcher/AggSchemaScanTest.java
+++ b/mondrian/src/it/java/mondrian/rolap/aggmatcher/AggSchemaScanTest.java
@@ -1,0 +1,130 @@
+/*
+ * This software is subject to the terms of the Eclipse Public License v1.0
+ * Agreement, available at the following URL:
+ * http://www.eclipse.org/legal/epl-v10.html.
+ * You must accept the terms of that agreement to use this software.
+ *
+ * Copyright (c) 2002-2019 Hitachi Vantara..  All rights reserved.
+ */
+
+package mondrian.rolap.aggmatcher;
+
+import junit.framework.Assert;
+import mondrian.olap.Util;
+import mondrian.rolap.RolapConnection;
+import mondrian.rolap.RolapConnectionProperties;
+import mondrian.test.FoodMartTestCase;
+
+import javax.sql.DataSource;
+import java.sql.Connection;
+import java.sql.DatabaseMetaData;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+/**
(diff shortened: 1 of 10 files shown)


mondrian/src/it/java/mondrian/rolap/aggmatcher/AggSchemaScanTest.java
```diff
@@ -0,0 +1,130 @@
...
+      sqlConnection = dataSource.getConnection();
+      DatabaseMetaData dbmeta = sqlConnection.getMetaData();
+      if ( !dbmeta.supportsSchemasInTableDefinitions() && !dbmeta.supportsCatalogsInTableDefinitions() ) {
+        System.out.println( "Database does not support schema or catalog in table definitions.  Cannot run test." );
+        return;
+      }
+      Util.PropertyList propertyList = new Util.PropertyList();
+      boolean foundSchema = false;
+      // Different databases treat catalogs and schemas differently.  Figure out whether foodmart is a schema or catalog in this database
+      if ( dbmeta.supportsSchemasInTableDefinitions() ) {
```
mkambol: It might make sense to first just try the sqlConnection.getSchema() and sqlConnection.getCatalog() calls, since for many databases these will actually give something back.  You'd want to wrap the calls in a try/catch since some other jdbc clients will throw an exception.

This could allow you to avoid the hard assumption that the schema or catalog will be named 'foodmart' for the databases that support one or both of those getXXX methods.

mondrian/src/main/java/mondrian/rolap/RolapSchema.java
```diff
@@ -300,7 +301,11 @@ protected Logger getLogger() {
      * @param catalogUrl URL of catalog
      * @param catalogStr Text of catalog, or null
      */
```
mkambol: Were you going to leave the original load() method, and add the new one as an overload?

cdeptula: Since the load method is protected and is not a public API, I was not concerned with maintaining the original load method.  I only overloaded methods that were public and we were changing the method signature.  I can overload the load method if you think it is necessary, but I personally do not see a reason to.

mkambol: It's possible that somebody has extended RolapSchema.  But I guess anybody who has customized to that extent could also adjust their implementation pretty easily to accommodate the new signature, so probably not a big deal either way.

cdeptula:
Closing PR and will create a new one after I run checkfile to fix style issues.

PR #1111, opened 2019-03-19
[BACKLOG-27566] - Added SLF4J-log4j bridge to schema workbench

graimundo:


nantunes:
R-6127480-1411745633-7522 jpeg

PR #1119, opened 2019-04-10
[BACKLOG-27905] Updated snowflake mondrian dialect to have snowflake …

ricksjames:
…native regex and pass more mondrian integration tests

### DIFF
--- a/demo/FoodMart.xml
+++ b/demo/FoodMart.xml
@@ -224,6 +224,9 @@ CONCAT(CONCAT("customer"."fname", ' '), "customer"."lname")
           <SQL dialect="teradata">
 "fname" || ' ' || "lname"
           </SQL>
+          <SQL dialect="snowflake">
+"customer"."fullname"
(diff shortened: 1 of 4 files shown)


mondrian/pom.xml
```diff
@@ -164,6 +164,12 @@
       <version>${mysql-connector-java.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
kurtwalker: Since we don't include all the other possible jdbc drivers, I think we should leave this out.  Readme suggests defining a profile in settings.xml.  I'd stick with that

mkambol:
Sorta surprised the Foodmart.xml changes don't break a bunch of tests.  Used to be there were hash checks in some tests.

mondrian/pom.xml
ricksjames: fixed - removed changes to pom.xml.

PR #1133, opened 2019-07-22
[PAD-147] - Cannot connect to Mondrian with SQL Server datasource using Integrated Security

bantonio82:
@pentaho/tatooine @pentaho-lmartins

### DIFF
--- a/mondrian/src/main/java/mondrian/rolap/RolapConnection.java
+++ b/mondrian/src/main/java/mondrian/rolap/RolapConnection.java
@@ -20,6 +20,7 @@ import mondrian.spi.*;
 import mondrian.spi.impl.JndiDataSourceResolver;
 import mondrian.util.*;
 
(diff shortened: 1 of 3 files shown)


mondrian/src/main/java/mondrian/rolap/RolapConnection.java
```diff
@@ -1134,6 +1133,22 @@ public boolean isWrapperFor( Class<?> iface ) throws SQLException {
     }
   }
 
+  private static String getJdbcConnectionString( Util.PropertyList connectInfo ) {
```
pentaho-lmartins: Can I see a unit test for this new method?

pentaho-lmartins: With / without DB name.
With / without integrated security.

PR #1151, opened 2019-08-23
[MONDRIAN-2674] Better logging of late SQL queries for segments.

lucboudreau:


lucboudreau:
@tkafalas For review.

PR #1157, opened 2019-10-10
[ANALYZER-3828] - Saving an inline model breaks localization

bcostahitachivantara:
@pentaho-lmartins 
@pentaho/tatooine 
@LeonardoCoelho71950 
@kurtwalker 
@bennychow 

I would like a review for this solution while i am trying write a integration test to this changes.


mondrian/src/main/java/mondrian/rolap/RolapSchemaPool.java
```diff
@@ -371,10 +372,43 @@ private static String getSchemaContent(
...
+        if (RolapSchema.LOGGER.isDebugEnabled()) {
+            RolapSchema.LOGGER.debug(
+              "Pool.get: translate schema \"" + catalogStr
+                + "\" using dynamic processor");
+        }
+        try {
+            final DynamicSchemaProcessor dynProc =
+              ClassResolver.INSTANCE.instantiateSafe(dynProcName);
+            if( dynProc instanceof LocalizingDynamicSchemaProcessor )
+                catalogStrProcessed = ((LocalizingDynamicSchemaProcessor)dynProc).filter( catalogStrProcessed, connectInfo);
```
lucboudreau: Do you really need to cast to localizing processor class? I understand that you created a new method in the localizing class, but you should only do this as a last resort. It breaks the API. The processor's API should be sufficient. You can use the filter method that already exists. There's no need to check for instanceof either.

https://github.com/pentaho/mondrian/blob/master/mondrian/src/main/java/mondrian/spi/DynamicSchemaProcessor.java#L53

mondrian/src/main/java/mondrian/rolap/RolapSchemaPool.java
```diff
@@ -371,10 +372,43 @@ private static String getSchemaContent(
...
+            // If exist a DynamicSchemaProcessor apply locale
+            if ( !Util.isEmpty( dynProcName ) ) {
+                return processDynamicSchemaWithCatalog( dynProcName, catalogStr, connectInfo );
+            }
+        }
 
         return catalogStr;
     }
 
+    private static String processDynamicSchemaWithCatalog(
```
lucboudreau: Can we refactor the code of processDynamicSchema() and make it more generic and merge its code with this new method? They largely do the same work and duplicate code.

mondrian/src/main/java/mondrian/i18n/LocalizingDynamicSchemaProcessor.java
```diff
@@ -109,7 +109,25 @@ public String filter(
         return schema;
     }
 
-    private String doRegExReplacements(String schema) {
+    public String filter( String schema, Util.PropertyList connectInfo ) {
```
lucboudreau: You shouldn't need to create a class specific method. You are also circumventing a required call to super.filter(). This also forces you to do class casts in weird places later below. I think you should use the existing filter method.

buildguy:
### :gear: Build finished in 4m 34s with command:
```console
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd \
-Daudit -P !assemblies -pl mondrian
```
:exclamation: No tests found!


mondrian/src/main/java/mondrian/i18n/LocalizingDynamicSchemaProcessor.java
bcostahitachivantara: With solutions above this method will be integrated into "api".

mondrian/src/main/java/mondrian/rolap/RolapSchemaPool.java
bcostahitachivantara: **Solution 1**
With the solution above, we can dismiss this method and call directly the new method:

`public String processSchema( String schema, Util.PropertyList connectInfo) throws Exception;` 

**Solution 2**
Instead call a different method, call only post-processing to apply translations.

mondrian/src/main/java/mondrian/rolap/RolapSchemaPool.java
bcostahitachivantara: The interface DynamicSchemaProcessor.java only have the method: 
`public String processSchema( String schemaUrl, Util.PropertyList connectInfo) throws Exception;` 

This method is implemented by FilterDynamicSchemaProcessor that it only creates an InputStream from schemaUrl.  

**Solution 1**
As solution, we can add a new method to interface to allow it to process schemas that has got from connectedInfo instead from a schemaUrl.

`public String processSchema( String schema, Util.PropertyList connectInfo) throws Exception;` 

Both of this methods have the same type of parameters, that means we cannot overload this method, we can add a extra parameter to allow differentiate or create a new method with a different name.

**Solution 2**
Another solution can be add a public method for pre-processing and another to post-processing.

lucboudreau: There's always solution 0. Use a stream as the API requires.

buildguy:
### :gear: Build finished in 4m 21s with command:
```console
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd \
-Daudit -P !assemblies -pl mondrian
```
:exclamation: No tests found!


buildguy:
### :gear: Build finished in 3m 52s with command:
```console
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd \
-Daudit -pl mondrian
```
:exclamation: No tests found!


mondrian/src/main/java/mondrian/spi/DynamicSchemaProcessor.java
```diff
@@ -50,9 +50,20 @@
...
+     * Modifies a Mondrian catalog.
+     *
+     * <p>An implementation should process catalog and apply dynamic changes
+     *
+     * @param catalog The URL of the catalog.
+     * @param connectInfo Connection properties.
+     * @return Returns the modified schema.
+     * @throws Exception if an error occurs.
+     */
+    public String processCatalog( String catalog, Util.PropertyList connectInfo) throws Exception;
```
lucboudreau: This method needs a default implementation in the interface. If not it will break all of the integrations by customers :)

https://www.baeldung.com/java-static-default-methods

The default should just return the catalog string as-is. It will also allow you to remove all of the new methods you put in the test classes.

buildguy:
### :gear: Build finished in 3m 36s with command:
```console
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd \
-Daudit -pl mondrian
```
:exclamation: No tests found!


buildguy:
### :gear: Build finished in 4m with command:
```console
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd \
-Daudit -pl mondrian
```
:exclamation: No tests found!


mondrian/src/it/java/mondrian/test/LocalizingDynamicSchemaProcessorTest.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Tests to make sure that our LocalizingDynamicSchemaProcessor works, with
+     * replacements. Does not test Mondrian is able to connect with the schema
+     * definition.
+     */
+    public void testProcessingCatalog() {
+        String variable = "%{translate}";
+        String translation = "Translate";
+        String catalog = "..." + variable + "...";
+        MondrianProperties.instance().LocalePropFile.set( "resources.properties" );
```
lucboudreau: Two things here:

 - Is there a way not to use external files? I don't particularly like the idea of an external test file. MAybe refactor a method into something you could override in the next line below when you create the LDSP. If not, at least put the file in a proper folder, not at the root of the test resources.

- The property needs to be reset at the end of this particular test method. Normally, all of our test are subclasses of FoodMartTest, which exposes the propSaver protected object. Setting properties through the propSaver will automatically reset them after the test. Might be worth looking into.

bcostahitachivantara: - I am going put a file in a different folder closer to class to be tested. I think that using a properties file inside a resource file (external file as you called it) is the better approach because it allow us to test loading file content to our object.

- I based my test on DynamicSchemaProcessorTest.java.

buildguy:
### :gear: Build finished in 4m 18s with command:
```console
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd \
-Daudit -pl mondrian
```
:exclamation: No tests found!


lucboudreau:
@bcostahitachivantara 

Sorry to bring this up now. I wish I had seen it earlier. Your test won't run. You need to add it to Main.java.

buildguy:
### :gear: Build finished in 3m 47s with command:
```console
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd \
-Daudit -pl mondrian
```
:exclamation: No tests found!

PR #1160, opened 2019-10-15
[MONDRIAN-2630] Mondrian: filter a set of members including null, plu…

LeonardoCoelho71950:
…s a numeric filter, results in incorrect data

Hey @mkambol, when a column wasn't constrained by a slicer, we weren't building the "IS NULL" expression, that's the reason why testNativeWithOverriddenNullMemberRepAndNullConstraint was failing.


buildguy:
### :gear: Build finished in 3m 25s with command:
```console
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd \
-Daudit -P !assemblies -pl mondrian
```
:exclamation: No tests found!

PR #1185, opened 2020-01-27
[MONDRIAN-2670] Regular Expression used in Analyzer's Find dialog not…

bennychow:
… passed down to query when using Vertica

### DIFF
--- a/mondrian/src/main/java/mondrian/spi/impl/VerticaDialect.java
+++ b/mondrian/src/main/java/mondrian/spi/impl/VerticaDialect.java
@@ -4,11 +4,9 @@
 // http://www.eclipse.org/legal/epl-v10.html.
 // You must accept the terms of that agreement to use this software.
 //
-// Copyright (c) 2002-2017 Hitachi Vantara..  All rights reserved.
+// Copyright (c) 2002-2020 Hitachi Vantara..  All rights reserved.
 */
 package mondrian.spi.impl;
-
-import mondrian.rolap.SqlStatement;
 
 import java.sql.Connection;
 import java.sql.ResultSetMetaData;
@@ -18,6 +16,12 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+
+import mondrian.rolap.SqlStatement;
+import mondrian.spi.DialectUtil;
 
 /**
  * Implementation of {@link mondrian.spi.Dialect} for the Vertica database.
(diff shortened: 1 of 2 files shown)


buildguy:
### :white_check_mark: Build finished in 8m
Build command:
```console
mvn clean verify -B -e -Daudit -amd -pl mondrian
```
---
 :exclamation: No tests found!
 ---
 
 

bennychow:
@pentaho/millenniumfalcon @lucboudreau @mkambol @kurtwalker Please help review and merge.  Thanks!

mondrian/src/main/java/mondrian/spi/impl/VerticaDialect.java
```diff
@@ -27,108 +31,165 @@
...
-
-    /**
-     * Creates a VerticaDialect.
-     *
-     * @param connection Connection
-     */
-    public VerticaDialect(Connection connection) throws SQLException {
-        super(connection);
-    }
+  private final String flagsRegexp = "^(\\(\\?([a-zA-Z]+)\\)).*$";
```
mkambol: It looks like this regex for finding the flags gets duplicated in other dialects.  Could the flag logic be extracted and moved either into JdbcDialectImpl or some utility class?

mkambol:
Looks good @bennychow.  Just one small comment about potentially extracting some duplicated logic.

mondrian/src/main/java/mondrian/spi/impl/VerticaDialect.java
bennychow: Thanks for reviewing Matt.  

It sure can.  I'm going to be working on getting regexp support for Redshift next so now is as good as time as ever to re-factor this a bit.

lucboudreau: Hey @bennychow I had this around my local box. I did some tests for Redshift and regular expressions. Hope this helps.

```
    @Override
    public boolean allowsRegularExpressionInWhereClause() {
        return true;
    }

    @Override
    public String generateRegularExpression(
        String source,
        String javaRegex)
    {
    	try {
            Pattern.compile(javaRegex);
        } catch (PatternSyntaxException e) {
            // Not a valid Java regex. Too risky to continue.
            return null;
        }

        // We might have to use case-insensitive matching
        javaRegex = DialectUtil.cleanUnicodeAwareCaseFlag(javaRegex);
        final Matcher flagsMatcher = flagsPattern.matcher(javaRegex);
        boolean caseSensitive = true;
        if (flagsMatcher.matches()) {
            final String flags = flagsMatcher.group(2);
            if (flags.contains("i")) {
                caseSensitive = false;
            }
        }
        if (flagsMatcher.matches()) {
            javaRegex =
                javaRegex.substring(0, flagsMatcher.start(1))
                + javaRegex.substring(flagsMatcher.end(1));
        }
        final Matcher escapeMatcher = escapePattern.matcher(javaRegex);
        while (escapeMatcher.find()) {
            javaRegex =
                javaRegex.replace(
                    escapeMatcher.group(1),
                    escapeMatcher.group(2));
        }
        final StringBuilder sb = new StringBuilder();

        // Now build the string.
        sb.append(source);
        sb.append(" IS NOT NULL AND ");
        
        // https://docs.aws.amazon.com/redshift/latest/dg/REGEXP_INSTR.html
        sb.append("REGEXP_INSTR(");
        sb.append(source);
        sb.append(",");
        quoteStringLiteral(sb, javaRegex);
        sb.append(",1,1,");
        sb.append(caseSensitive ? "c" : "i");
        sb.append(") > 0");

        return sb.toString();
    }
```

bennychow: Thanks Luc!  This will definitely be helpful.

PR #1186, opened 2020-01-29
[MONDRIAN-2670] Add support for regular expressions in Vertica and Re…

bennychow:
…dshift dialect.  Re-factor flag mapping logic to shared JdbcDialectImpl.


bennychow:
@mkambol @lucboudreau @kurtwalker Hi Guys, could you review again? I have refactored the flag matching logic.  Using Analyzer's Contains filter, I tested the regular expression push down on Vertica, Redshift, Mysql and Postgresql and they all produced the expected SQL and results.

@pentaho/millenniumfalcon

PR #1191, opened 2020-02-24
Separated CLEANUP commit from functional commit

mkambol:
Reviewing MONDRIAN-2662 will be much easier in its separate commit.


buildguy:
### :white_check_mark: Build finished in 8m 52s
Build command:
```console
mvn clean verify -B -e -Daudit -amd
```
---
 :exclamation: No tests found!
 ---
 
 

buildguy:
### :white_check_mark: Build finished in 7m 58s
Build command:
```console
mvn clean verify -B -e -Daudit -amd -pl mondrian
```
---
 :exclamation: No tests found!
 ---
 
 

buildguy:
### :white_check_mark: Build finished in 9m 6s
Build command:
```console
mvn clean verify -B -e -Daudit -amd
```
---
 :exclamation: No tests found!
 ---
 
 

buildguy:
### :white_check_mark: Build finished in 9m 6s
Build command:
```console
mvn clean verify -B -e -Daudit -amd
```
---
 :exclamation: No tests found!
 ---
 
 

mondrian/src/main/java/mondrian/calc/impl/ArrayTupleList.java
```diff
@@ -24,6 +25,7 @@
 {
     private transient Member[] objectData;
     private int size;
+    private static final int cjMaxSize = MondrianProperties.instance().ResultLimit.get();
```
bennychow: @mkambol Should _**static**_ be removed?  It is possible to re-load the Mondrian properties at run-time.  I checked all current uses of this ResultLimit property and nowhere else saves it into a static location.

mkambol: Good catch!

mondrian/src/main/java/mondrian/calc/impl/ArrayTupleList.java
```diff
@@ -247,6 +249,11 @@ private void ensureCapacity(int minCapacity) {
             // Up to next multiple of arity.
             final int rem = newCapacity % arity;
             newCapacity += (arity - rem);
+            if (cjMaxSize > 0 && newCapacity > cjMaxSize) {
```
bennychow: @mkambol I don't think checking the newCapacity against the limit is the right thing to do.  Let's assume the limit is set to 1M.  Suppose you have a list of 700K tuples and you need to just add 10 more tuples to this list.  Lets also suppose that the array's capacity was exactly 700K and now we need to increase it by 1.5x to accommodate the additional 10.  However, as soon as we increase the capacity of the array to 700K+350K=1.05M, this will hit the 1M limit and error out.  As a user, I know my query should have only returned 700,010 tuples which is well under the 1M limit.

I think the right fix would be to cap the newCapacity at the limit and only error out if the array was filled beyond this limit.

PR #1200, opened 2020-03-11
Revert "[SP-5453] Backport of MONDRIAN-2670 - Regular Expression used in Analyzer's Find dialog not passed down to query when using Vertica & Redshift (9.0 Suite)"

dcleao:
Reverts pentaho/mondrian#1193

Reverting due to being considered high risk.


dcleao:
Ended up resetting history to clean up the mess.

PR #1233, opened 2020-06-25
[MONDRIAN-2689] Regression with CurrentMember MDX function performanc…

bennychow:
…e with large slicers

This fix primarily improves the performance of the CurrentMember MDX function when it tries to validate that the function is not referencing a slicer that contains more than one member from the hierarchy.  This is done by using a cached map to lookup the slicer members for a given hierarchy on the slicer axis instead of doing an expensive iteration over all slicer members at the point of validation.

I also tried to improve the performance of large slicers in general and found that RolapResult.AxisMemberList.members was using an inefficient data structure.  There's no point in using a ConcatenableList there because members are always added one by one and not as lists of members.  If we used a Set instead of List, this would be even better since it would make duplicate checking faster here:

https://github.com/pentaho/mondrian/blob/master/mondrian/src/main/java/mondrian/rolap/RolapResult.java#L1635

However, this would cause unit tests to break because the generated SQL would be non-deterministic because the order of items in sets are not guaranteed.

@mkambol @lucboudreau  @pentaho/ackbar Please help review this performance fix.  Thanks

### DIFF
--- a/mondrian/src/main/java/mondrian/rolap/agg/SegmentLoader.java
+++ b/mondrian/src/main/java/mondrian/rolap/agg/SegmentLoader.java
@@ -342,6 +342,7 @@ public class SegmentLoader {
                 case INT:
                 case LONG:
                 case DOUBLE:
+                case DECIMAL:
                     Object o = rows.getObject(j);
                     if (useGroupingSet
                         && (o == null || o == RolapUtil.sqlNullValue)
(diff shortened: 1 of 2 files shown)


buildguy:
### :white_check_mark: Build finished in 16m 42s
Build command:
```console
mvn clean verify -B -e -Daudit -amd -DrunITs -pl mondrian
```
---
:ok_hand: All tests passed!
 
Tests run: **3705**, Failures: **0**, Skipped: **0** &nbsp;&nbsp; [Test Results](http://wingman.pentaho.net/job/projects/job/mondrian/job/PR-1233/1/testReport)

 ---
 
 

lucboudreau:
@bennychow How about using a SortedSet / TreeSet? It'll keep the natural ordering if you don't pass a comparator. Not sure how it performs in this specific scenario but it might work.

https://docs.oracle.com/javase/7/docs/api/java/util/TreeSet.html

bennychow:
@lucboudreau Great suggestion.  I tried using TreeSet instead of ArrayList and used an even larger slicer and the performance improvement was very significant.

bennychow:
Ugh.. can't compare members when running CompoundSlicerTest:

Caused by: java.lang.ClassCastException: mondrian.rolap.RolapBaseCubeMeasure cannot be cast to mondrian.rolap.RolapCubeMember
	at mondrian.rolap.RolapCubeMember.compareTo(RolapCubeMember.java:95)
	at java.util.TreeMap.getEntry(TreeMap.java:352)
	at java.util.TreeMap.containsKey(TreeMap.java:232)
	at java.util.TreeSet.contains(TreeSet.java:234)
	at mondrian.rolap.RolapResult$AxisMemberList.mergeMember(RolapResult.java:1636)
	at mondrian.rolap.RolapResult$AxisMemberList.mergeTuple(RolapResult.java:1628)
	at mondrian.rolap.RolapResult$AxisMemberList.mergeTupleIter(RolapResult.java:1611)
	at mondrian.rolap.RolapResult$AxisMemberList.mergeTupleList(RolapResult.java:1602)
	at mondrian.rolap.RolapResult.executeAxis(RolapResult.java:1020)
	at mondrian.rolap.RolapResult.evalLoad(RolapResult.java:842)
	at mondrian.rolap.RolapResult.loadMembers(RolapResult.java:799)
	at mondrian.rolap.RolapResult.<init>(RolapResult.java:296)
	at mondrian.rolap.RolapConnection.executeInternal(RolapConnection.java:651)
	... 7 more

The performance improvement is pretty significant.  For example, my test case has 20,000 slicer members and the performance of the different data structures are as follows:

25 seconds with HashSet
26 seconds with TreeSet
33 seconds with ArrayList
41 seconds with ConcatenableList

Any other suggestions?  Maybe use both ArrayList and HashSet data structures at the same time?

PR #1296, opened 2022-01-07
[BISERVER-14733] Use same rollover strategy and file names as origina…

bennychow:
…l log4j.xml


bennychow:
@lucboudreau  Could you review and merge?  This is just making the log4j2.xml work exactly the same as the original log4j.xml in terms of roll over strategy and file names.

bennychow:
@lucboudreau Could you review and merge?  This just makes the log4j2.xml work mostly the same as the original log4j.xml.

PR #1314, opened 2022-01-27
[BISERVER-14720] Add log4j-jcl dependency

peterrinehart:


assemblies/psw-ce/src/main/resources/log4j2.xml
```diff
@@ -35,6 +34,7 @@
         <!-- Append messages to the console -->
         <!-- ============================== -->
         <Console name="CONSOLE">
+            <ThresholdFilter level="INFO" onMatch="ACCEPT" onMismatch="DENY"/>
```
bennychow, line 37: I don't think you need to specify any attributes... the default for onMatch is NEUTRAL.

PR #1396, opened 2024-09-19
fix: fix repeated formula element

hugo-jose:
Fix incorrect xsd definition that lead to the following error:

[ERROR] Error while parsing schema(s).Location [ file:/Users/hjose/_work/code/semantic-model-editor/backend/src/main/resources/xsd/mondrian.xsd{1723,69}].
com.sun.istack.SAXParseException2: Property "Formula" is already defined. Use &lt;jaxb:property> to resolve this conflict.


hugo-jose:
PR is not valid. I will close it.

PR #1467, opened 2025-09-26
[BACKLOG-43902]-Java 21 : compatibility check for Pentaho and fix Build and Run time issues while building / testing

SahithiKommagani6:
This pull request makes a minor change to the build process by updating the `mondrian/build.xml` file. The main change is switching the `fork` attribute from "no" to "true" for a Java execution task, which will cause the Java process to run in a separate JVM.
Forking starts the Java process separately, allowing better compatibility with newer JVM options

### DIFF
--- a/mondrian/build.xml
+++ b/mondrian/build.xml
@@ -69,7 +69,7 @@
 
     <java classpath="${classpath}"
         classname="mondrian.util.PropertyUtil"
-        fork="no">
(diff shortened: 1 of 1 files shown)


krishnaalluhitachi (approved):

lucboudreau:
The build does not use build.xml anymore. That's an old artifact.

